PR #13, opened 2019-01-24
Feature/trial events

nwingt:


nnkken (approved):

src/routes/civic/index.js
```diff
@@ -133,4 +134,109 @@ router.get('/csonline', async (req, res, next) => {
...
+      .collection('trialEvents')
+      .doc(id)
+      .get();
+
+    if (doc.exists) {
+      const { start, end } = doc.data();
+      const now = Date.now();
+      if (now < start) {
+        res.sendStatus(404);
+      } else if (now > end) {
```
williamchong: Use `return` like that api below for easier reading and consistent style?
```
    if (now < start) {
      res.sendStatus(404);
      return;
    }
```

williamchong (changes requested):

src/routes/civic/index.js
```diff
@@ -133,4 +134,113 @@ router.get('/csonline', async (req, res, next) => {
...
+    }
+    if (now > end) {
+      res.sendStatus(410);
+      return;
+    }
+
+    const trialEnd = new Date(now);
+    trialEnd.setMonth(trialEnd.getMonth() + 1);
+    const createObj = {
+      civicLikerStatus: 'subscribed',
```
mckingho: suggest the createObj = {
  currentPeriodStart: now,
  currentType: 'trial',
}
A trigger function will update other details.

mckingho (changes requested):

mckingho:
Do we have trial quota?

williamchong:
*Add a `quota` field for each event, which we can set a number higher than the expected users
just in case we got surprised spam

src/routes/civic/index.js
```diff
@@ -213,7 +224,10 @@ router.post('/trial/events/:eventId/join', jwtAuth('write'), async (req, res, ne
       currentPeriodStart: now,
       currentPeriodEnd: trialEnd.getTime(),
     };
-    await subscriptionDbRef.doc(userId).create(createObj);
+    await db.runTransaction(async (t) => {
+      await t.update(trialEventRef, { regCount: regCount + 1 });
```
williamchong, line 219: `regCount` should be `.get()` inside this transaction

williamchong (changes requested):

williamchong (approved):

PR #15, opened 2019-01-30
✨ Add Civic Liker registration quota check API

nwingt:


src/routes/civic/index.js
```diff
@@ -16,19 +16,33 @@ import {
...
+    const configDoc = await configRef.doc('civicLiker').get();
+    if (configDoc.exists) {
+      const { regCount, regQuota } = configDoc.data();
+      if (regCount < regQuota) {
+        res.sendStatus(200);
+        return;
+      }
+    }
+
+    res.sendStatus(404);
```
williamchong: @nnkken what do you think about this 404

nnkken: Should it return `404` when `regCount >= regQuota`? Sounds weird...

nnkken (approved):

williamchong:
... we probably should just make 1 `civicLikerStatus` field will all those `trial` `expired` `active` and stuff?

PR #31, opened 2019-04-04
✨ Support `transferMultiple`

Aludirk:
📦 Update `firebase-admin` & `sharp`
✨ Support test on specific file by `yarn test ${test_file}`
✨ Support `transferMultiple` for tx info and tx history with address filter


src/util/ValidationHelper.js
```diff
@@ -125,27 +125,47 @@ export function filterTxData({
...
   value,
   status,
   type,
   remarks,
   httpReferrer,
   completeTs,
   ts,
-}) {
-  return {
+}, filterAddr) {
```
nwingt: Prefer using object for options e.g. `options = { filterAddr: '' } `

src/routes/tx/index.js
```diff
@@ -20,10 +20,11 @@ const router = Router();
 router.get('/id/:id', async (req, res, next) => {
   try {
     const txHash = req.params.id;
+    const filterAddress = req.query.address;
```
williamchong:
```suggestion
  const { address: filterAddress } = req.query;
```

src/routes/tx/index.js
```diff
@@ -10,29 +11,64 @@ import { jwtAuth } from '../../middleware/jwt';
...
   filterTxData,
-  filterMultipleTxData,
   checkAddressValid,
 } from '../../util/ValidationHelper';
 
 const web3Utils = require('web3-utils');
 
 const router = Router();
 
+function filterMultipleTxData({
```
williamchong: `normalizeMultipleTxData` should be better since without `opt` this does not filter anything

src/routes/tx/index.js
```diff
@@ -10,29 +11,64 @@ import { jwtAuth } from '../../middleware/jwt';
...
+  from,
+  to,
+  toIds,
+  value,
+}, opt) {
```
williamchong: Rename `opt` -> `filter` sounds more verbose

src/routes/tx/index.js
```diff
@@ -10,29 +11,64 @@ import { jwtAuth } from '../../middleware/jwt';
...
+  to.forEach((addr, index) => {
+    if (addr === address) {
+      if (tos.length === 0) {
+        tos.push(addr);
+        ids.push(toIds[index]);
+        values.push(value[index]);
+      } else {
+        if (tos[0] !== addr) throw new Error('Filter address is not matched');
+        if (ids[0] !== toIds[index]) throw new Error('Filter ID is not matched');
+        values[0] = new BigNumber(values[0]).plus(new BigNumber(value[index])).toString();
```
williamchong: using `Array` for reduce operation is confusing, should either stick with single value or use map

src/routes/tx/index.js
```diff
@@ -10,29 +11,64 @@ import { jwtAuth } from '../../middleware/jwt';
...
     if (doc.exists) {
       const payload = doc.data();
-      if (Array.isArray(payload.to)) {
-        res.json(filterMultipleTxData(payload, filterAddress));
-      } else {
-        res.json(filterTxData(payload));
-      }
-    } else {
-      res.sendStatus(404);
+      Object.assign(payload, filterMultipleTxData(payload, { address }));
```
williamchong: FYI this can also be written in spread 
```const payload = {...payload, ...filterMultipleTxData(payload, { address })}```

williamchong (changes requested):

williamchong (approved):

PR #51, opened 2019-08-21
🍏 Update isSafari check

nwingt:


williamchong:
will this include ios chrome? can we exclude it somehow? Samesite:none is mainly used to mitigate chrome's 3rd party cookie policy

nwingt:
@williamchong007 Updated

williamchong (approved):

nnkken (approved):

PR #87, opened 2019-11-01
✨ Add LikeCoin price api

mckingho:


src/routes/misc/price.js
```diff
@@ -0,0 +1,33 @@
...
+
+router.get('/price', async (req, res) => {
+  let price = LIKE_DEFAULT_PRICE;
+  try {
+    const prices = await Promise.all([
+      axios.get(COINGECKO_PRICE_URL)
+        .then(r => r.data.market_data.current_price.usd)
+        .catch(() => LIKE_DEFAULT_PRICE),
+      axios.get(COINMARKETCAP_PRICE_URL)
+        .then(r => parseFloat(r.data[0].price_usd))
```
williamchong: plz support currency override via querystring (usd as default if not set)

williamchong (changes requested):

src/routes/misc/price.js
```diff
@@ -0,0 +1,34 @@
...
+  COINMARKETCAP_PRICE_URL,
+  LIKE_DEFAULT_PRICE,
+} from '../../constant';
+
+const router = Router();
+
+const CACHE_IN_S = 30;
+
+router.get('/price', async (req, res) => {
+  const { convert = 'usd' } = req.query;
```
williamchong: `currency` should be a better name

williamchong (changes requested):

nnkken (approved):

src/routes/misc/price.js
```diff
@@ -0,0 +1,40 @@
...
+        },
+      }).then(r => parseFloat(r.data.LIKE.quote[currency].price))
+        .catch(() => 0),
+    ]);
+    price = Math.max(...prices);
+  } catch (err) {
+    console.error(err);
+  }
+  res.set('Cache-Control', `public, max-age=${CACHE_IN_S}, s-maxage=${CACHE_IN_S}, stale-if-error=${CACHE_IN_S}`);
+  res.send({ price });
```
mckingho, line 47: @nnkken Should I return `price:null` or `status 500` for error case?

nnkken (approved):

src/routes/misc/price.js
```diff
@@ -0,0 +1,49 @@
...
+          'X-CMC_PRO_API_KEY': CMC_PRO_API_KEY,
+        },
+      }).then(r => parseFloat(r.data.LIKE.quote[currency].price))
+        .catch(() => undefined),
+    ]);
+    if (prices.every(p => p === undefined)) {
+      if (currency !== 'usd') throw new Error('Failed to get price.');
+      price = LIKE_DEFAULT_PRICE;
+    } else {
+      price = Math.max(...prices.filter(p => p !== undefined));
```
williamchong: instead of checking undefined,  check for `isNaN`.
Since we need to iterate the array in both line 32 and 36 anyway, we can just filter the array by `isNaN` in line 32, and use the resulting array's length for the `if` clause. The result array can be directly use in line 36.

williamchong (changes requested):

PR #94, opened 2019-11-27
🐛 Use authcore user access token to fetch wallet instead of service a…

williamchong:
…ccount

### DIFF
--- a/src/routes/users/platforms/index.js
+++ b/src/routes/users/platforms/index.js
@@ -8,11 +8,11 @@ import {
   tryToUnlinkOAuthLogin,
 } from '../../../util/api/users';
 import { fetchMattersOAuthInfo, fetchMattersUser } from '../../../util/oauth/matters';
-import { createAuthCoreCosmosWalletIfNotExist } from '../../../util/authcore';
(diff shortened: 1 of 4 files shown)


src/routes/users/platforms/index.js
```diff
@@ -96,7 +96,7 @@ router.post('/login/:platform/add', jwtAuth('write'), async (req, res, next) =>
     let platformUserId;
     switch (platform) {
       case 'authcore': {
-        const { idToken } = req.body;
+        const { idToken, accessToken } = req.body;
```
nnkken, line 99: should guard existence of fields?

src/routes/users/registerLogin.js
```diff
@@ -81,7 +81,11 @@ router.post(
 
       switch (platform) {
         case 'authcore': {
-          const { idToken, ...authCorePayload } = req.body;
+          const {
+            idToken,
+            accessToken,
+            ...authCorePayload
+          } = req.body;
```
nnkken, line 88: should also guard existence of `accessToken`?

nnkken (approved):

PR #97, opened 2019-12-01
💩 Ignore csrf checking for liker land app

williamchong:


williamchong:
@nwingt I tested on android app on my phone (should be in internal test v0.11.2) and ios master dev build via simulator
Haven't test on prod iOS device but the above 2 cases seems to hint the user agent pattern is `LikeCoinApp-(platform)`

nwingt (approved):

PR #127, opened 2020-02-06
Add mediaChannel field to user info

mckingho:


nnkken (approved):

williamchong:
not pass lint, also if it is an array, how about `mediaChannels` with s?

PR #128, opened 2020-02-14
✨ Add avatar from url when register

williamchong:
tested on dev servers, should be working no problem

### DIFF
--- a/src/util/fileupload.js
+++ b/src/util/fileupload.js
@@ -1,3 +1,4 @@
+import axios from 'axios';
 import {
   bucket as fbBucket,
 } from './firebase';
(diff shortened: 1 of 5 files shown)


src/util/fileupload.js
```diff
@@ -54,4 +55,27 @@ export async function handleAvatarUploadAndGetURL(user, file, avatarSHA256) {
...
+export async function handleAvatarLinkAndGetURL(user, url) {
+  const { data } = await axios.get(url, {
+    responseType: 'arraybuffer',
+    timeout: 5000,
+  });
+  const buffer = new Uint8Array(data);
+  const type = fileType(buffer);
+  if (!SUPPORTED_AVATER_TYPE.has(type && type.ext)) {
+    console.error(`unsupported file format! ${(type || {}).ext || JSON.stringify(type)}`);
+    return undefined;
```
nnkken: Should it throw error and let caller handle it?

src/util/fileupload.js
```diff
@@ -54,4 +55,27 @@ export async function handleAvatarUploadAndGetURL(user, file, avatarSHA256) {
...
+  }
+  const resizedBuffer = await sharp(data).resize(400, 400).toBuffer();
+  const filename = url.split('/').pop();
+  const file = {
+    buffer: resizedBuffer,
+    filename,
+    originalname: filename,
```
nnkken: seems unnecessary since you defined newFilename for `uploadFileAndGetLink`?

src/util/fileupload.js
```diff
@@ -54,4 +55,27 @@ export async function handleAvatarUploadAndGetURL(user, file, avatarSHA256) {
...
 }
 
+export async function handleAvatarLinkAndGetURL(user, url) {
+  const { data } = await axios.get(url, {
+    responseType: 'arraybuffer',
+    timeout: 5000,
+  });
+  const buffer = new Uint8Array(data);
+  const type = fileType(buffer);
+  if (!SUPPORTED_AVATER_TYPE.has(type && type.ext)) {
```
nwingt: `AVATER`

src/util/fileupload.js
williamchong: I think this was done to mitigate invalid avatar blocking user registration a while ago

nnkken (approved):

PR #148, opened 2020-05-29
✨ Add users notification api

williamchong:


nwingt (commented):
I am not sure whether notification feature is app only, should name the collection `appNotification` more generic?

williamchong:
Should the api path still be under `/app` in this case?

nwingt (approved):

PR #156, opened 2020-06-15
✨ Add bookmarks and follow api

williamchong:


src/routes/users/bookmarks.js
```diff
@@ -0,0 +1,122 @@
+import { Router } from 'express';
+import { userCollection as dbRef } from '../../util/firebase';
+import { filterBookmarks } from '../../util/ValidationHelper';
+import { jwtAuth } from '../../middleware/jwt';
+import { addUrlToMetadataCrawler } from '../../util/api/users/bookmarks';
+
+const uuidv4 = require('uuid/v4');
+const parse = require('url-parse');
```
nwingt: `parse` is a bit confusing, how about `urlParse`?

src/routes/users/bookmarks.js
```diff
@@ -0,0 +1,122 @@
...
+      return;
+    }
+    const query = await dbRef
+      .doc(user)
+      .collection('bookmarks')
+      .where('url', '==', url)
+      .limit(1)
+      .get();
+    if (query.docs.length) {
+      res.status(400).send('INVALID_URL');
```
nwingt: Better to return status code 409, and something like `BOOKMARK_EXISTS` as error code

src/routes/users/bookmarks.js
```diff
@@ -0,0 +1,122 @@
...
+    let targetRef;
+    if (!bookmarkID) {
+      const query = await dbRef
+        .doc(user)
+        .collection('bookmarks')
+        .where('url', '==', url)
+        .limit(1)
+        .get();
+      if (!query.docs || !query.docs.length) {
+        res.status(404).send('URL_NOT_FOUND');
```
nwingt: BOOKMARK_NOT_FOUND is better?

src/routes/users/bookmarks.js
```diff
@@ -0,0 +1,122 @@
...
+      }
+      targetRef = query.docs[0].ref;
+    } else {
+      const bookmarkRef = dbRef
+        .doc(user)
+        .collection('bookmarks')
+        .doc(bookmarkID);
+      const bookmarkDoc = await bookmarkRef.get();
+      if (!bookmarkDoc.exists) {
+        res.status(404).send('URL_NOT_FOUND');
```
nwingt: Same as above

src/routes/users/follow.js
```diff
@@ -0,0 +1,71 @@
...
+      .set({
+        isFollowed: false,
+        ts: Date.now(),
+      }, { merge: true });
+    res.sendStatus(200);
+  } catch (err) {
+    next(err);
+  }
+});
+
```
nwingt, line 70: Not need to handle 404?

nwingt (changes requested):

src/routes/users/follow.js
williamchong: To mimic the behaviour of liker.land api, `DELETE` endpoint can be used to unfollow a user which was not followed in the db. This is also why I used `.set( , { merge: true )`

nwingt (approved):

PR #162, opened 2020-06-23
✨ Add mobile device id logging for app

williamchong:


nwingt (commented):
Have u considered refactoring the deviceId payload injection?

williamchong:
Maybe we will need some kind of client side signing in the future, but currently it is like user agent, incoming request can freely mutate it and we use it for statistic purpose now.

nnkken (approved):

PR #169, opened 2020-06-30
💩 Try to fix firestore data not plain object error

williamchong:


nnkken (approved):
I don't think this will fix the error, but it should be harmless to try

nwingt (approved):

PR #171, opened 2020-07-02
💩 Try to log referrer api error data

williamchong:


nwingt (commented):
Is this for locating the error?

williamchong:
yes, also added a minor `toString()`

nnkken (approved):

PR #172, opened 2020-07-02
🐛 Fix authcore contact verification field format

williamchong:


nnkken (approved):

src/util/authcore.js
```diff
@@ -26,13 +26,19 @@ export async function getAuthCoreUser(accessToken) {
     primary_email: email,
     primary_email_verified: emailVerifiedTs,
   } = data;
+  let isEmailVerified = false;
+  if (emailVerifiedTs && typeof emailVerifiedTs === 'string') {
+    isEmailVerified = emailVerifiedTs && (new Date(emailVerifiedTs)).getTime() > 0;
```
nwingt, line 31: can remove `emailVerifiedTs &&` extra check

williamchong: actually i should remove the one in if condition to make the structure simpler

nnkken (approved):

nwingt (approved):

PR #175, opened 2020-07-09
🥅 Catch avatar error to prevent crash

williamchong:


src/routes/users/registerLogin.js
```diff
@@ -224,7 +224,13 @@ router.post(
       const { file } = req;
       let avatarUrl;
       if (file) {
-        avatarUrl = await handleAvatarUploadAndGetURL(user, file, avatarSHA256);
+        try {
+          avatarUrl = await handleAvatarUploadAndGetURL(user, file, avatarSHA256);
+        } catch (err) {
+          console.error('Avatar file handling error:');
+          console.error(err);
+          throw new ValidationError('INVALID_AVATAR');
```
nwingt, line 232: Can we use the placeholder without throwing error?

nwingt: OK, it is user update. The filename is really fake...

nwingt (approved):

src/routes/users/registerLogin.js
williamchong: Maybe can rename to better name, now this file contains register login update and sync authcore, all related to user profile data manipulation

PR #176, opened 2020-07-09
✨ Add get bookmark API

nwingt:


src/routes/users/bookmarks.js
```diff
@@ -9,7 +9,51 @@ const urlParse = require('url-parse');
...
+    let doc;
+    const { user } = req.user;
+    if (!bookmarkID) {
+      const qs = await dbRef
+        .doc(user)
+        .collection('bookmarks')
+        .where('url', '==', url)
+        .limit(1)
+        .get();
+      [doc] = qs.docs;
```
williamchong: This part should be put inside the `if (url)` block above?

src/routes/users/bookmarks.js
```diff
@@ -9,7 +9,51 @@ const urlParse = require('url-parse');
 
 const router = Router();
 
-router.get('/bookmarks', jwtAuth('read:bookmarks'), async (req, res, next) => {
+router.get('/bookmarks/:id?', jwtAuth('read:bookmarks'), async (req, res, next) => {
```
williamchong: If linter does not complain, `async (req, res, next)` should be on a new line
We can then add a line of comment before each middleware, stating what case does each of them handle

williamchong (changes requested):

williamchong (approved):
Maybe we should guard the case bookmarkID and url both exists with HTTP 400, to avoid confusion

williamchong (changes requested):
Be more explicit with error msg, e.g. `CANNOT_DEFINE_BOTH_URL_AND_ID`?

williamchong (approved):

PR #177, opened 2020-07-10
Add getting follow status

nwingt:


src/routes/users/follow.js
```diff
@@ -35,13 +35,36 @@ router.get('/follow/users', jwtAuth('read:follow'), async (req, res, next) => {
...
+  try {
+    const { user } = req.user;
+    const { id } = req.params;
+    const doc = await dbRef
+      .doc(user)
+      .collection('follow')
+      .doc(id)
+      .get();
+    if (!doc.exists) {
+      res.status(204).send('NO_FOLLOW_RECORD');
```
williamchong: Not sure should be 404 or 204

nwingt: Summon @nnkken for the judge

nnkken: Vote for 404
But if exception in front end code in this case is weird, then I prefer 200 with empty / `null` response

williamchong (approved):

PR #182, opened 2020-07-20
♻️ Refactor civic trial handler to external webhook

williamchong:


src/routes/civic/index.js
```diff
@@ -52,93 +46,25 @@ router.get('/trial/events/:id', async (req, res, next) => {
...
-      referrer,
-      locale,
-      registerTime,
-      mediaChannels,
-    });
+    const { eventId } = req.params;
+    const payload = req.body || {};
+    const response = await axios.post(
+      CIVIC_LIKER_TRIAL_WEBHOOK,
+      { ...payload || eventId },
```
nwingt: Seems even `payload` is undefined, `eventId` is not inside the post data
how about `payload || { eventId }`

nwingt (changes requested):

src/routes/civic/index.js
williamchong: I actually meant to write `{ ...payload, eventId }`
Will change

nwingt (approved):

PR #185, opened 2020-08-10
✨ Update notification API

nwingt:


src/routes/users/notifications.js
```diff
@@ -25,6 +51,38 @@ router.get('/notifications', jwtAuth('read:notifications'), async (req, res, nex
...
+      .where('isRead', '==', false)
+      .orderBy('ts', 'desc')
+      .startAfter(before)
+      .get();
+    if (query.docs.length) {
+      const batch = db.batch();
+      query.docs.forEach((d) => {
+        batch.update(d.ref, { isRead: true });
+      });
+      await batch.commit();
```
williamchong: Batch has a limit of 250 or 500 write per batch, a range request like this might hit this limit.
Please commit once per 250/500 limit (probably using a counter and `%`) , and rmb to commit the remaining in the end of request

williamchong (changes requested):

src/routes/users/notifications.js
```diff
@@ -8,13 +8,39 @@ const router = Router();
...
+        queryRef = queryRef.startAfter(before);
+      } catch (err) {
+        // no-op
+      }
+    }
+    if (limit) {
+      try {
+        limit = Number(limit);
+      } catch (err) {
+        // no-op
```
williamchong, line 39: if limit is a non-numerical value, and it errors here, does it break the checking below with non-numerical true-ful value?
Maybe should split input `limit` and value `Number(limit)` into 2 values

src/routes/users/notifications.js
```diff
@@ -25,6 +51,44 @@ router.get('/notifications', jwtAuth('read:notifications'), async (req, res, nex
...
 
+router.post('/notifications/read', jwtAuth('write:notifications'), async (req, res, next) => {
+  try {
+    const { user } = req.user;
+    let { before } = req.query;
+    if (before) {
+      try {
+        before = Number(before);
+      } catch (err) {
+        res.status(400).send('MISSING_BEFORE');
```
williamchong: `INVALID_BEFORE` since invalid values fall into this catch also?

williamchong (changes requested):

src/routes/users/notifications.js
nwingt: I don't think there is any truthy value while also non-numerical that return by `Number()`. You said split 2 values, you mean variables or what?

williamchong (approved):

PR #190, opened 2020-08-27
✨ Add archive bookmark API

williamchong:
Add default isArchived flag for bookmark entries to prepare for data migration.
Add querystring `archived` in GET API, which has default value 0. `archived=1` will return only archived bookmarks, while `archived=` (empty value) will return both archived and unarchived bookmark.


nwingt (approved):
I think pagination will be needed

PR #194, opened 2020-09-08
✨ Add pagination for bookmark and follow

williamchong:


williamchong:
https://github.com/likecoin/likecoin-api-public/pull/200

PR #213, opened 2021-01-07
✨ Add preferences API

nwingt:


williamchong (approved):
See if you want to add some trivial test cases

nwingt:
Added tests

williamchong (approved):

PR #217, opened 2021-01-29
🐛 Fix wrong user query param and unify all authCoreUserId casing

williamchong:


nnkken (approved):

nwingt (approved):
Anyhow, `Authcore` is the correct casing

PR #221, opened 2021-02-08
 🔒 Split likecoin and likecoin_button cookie and add ES256 jwt

williamchong:
Change to ES256 to shorten the cookie
Currently we still support active session of RS256, which we can remove later


nnkken (approved):

src/util/jwt.js
```diff
@@ -12,35 +12,75 @@ const config = require('../../config/config');
...
+  try {
+    const es256secret = fs.readFileSync(secretECDSACertPath);
+    if (es256secret) {
+      signAlgorithms.push('ES256');
+      signSecrets.ES256 = es256secret;
+      if (!internalDefaultSignAlgorithm) internalDefaultSignAlgorithm = 'ES256';
+    }
+  } catch (err) {
+    console.error(err);
+    console.error('RSA sign key not exist for jwt');
```
nnkken: `ECDSA` not `RSA`

PR #222, opened 2021-02-19
✨ Add creator pitch API

nwingt:


src/routes/users/preferences.js
```diff
@@ -23,12 +23,38 @@ router.get('/preferences', jwtAuth('read:preferences'), async (req, res, next) =
...
+          break;
+        }
+      }
+      payload.creatorPitch = creatorPitch.slice(0, i);
+    }
+
+    if (Object.keys(payload)) {
+      await dbRef.doc(user).set(payload, { merge: true });
     }
-    await dbRef.doc(user).update({ locale });
```
williamchong: User document should always exists in this case?
If we have assumption on it might not exist, we should check for user doc exists in the beginning of this route.

williamchong (changes requested):

src/util/ValidationHelper.js
```diff
@@ -102,6 +105,7 @@ export function filterUserDataMin(userObject, types = []) {
     isCivicLikerTrial,
     isSubscribedCivicLiker,
     civicLikerSince,
+    creatorPitch,
   };
   if (types.includes('payment')) {
```
williamchong, line 108: Would you consider using `types` querystring so we only include `creatorPitch` in response only when it is needed?

williamchong (approved):

PR #225, opened 2021-03-11
🥅Catch bad update payload

WeiJun0827:


src/routes/users/registerLogin.js
```diff
@@ -212,6 +212,10 @@ router.post(
         locale,
       } = req.body;
       let { isEmailEnabled } = req.body;
+      if (!displayName && !locale) {
+        res.sendStatus(400);
+        return;
+      }
```
williamchong: Instead of hardcoding here, can we check the `updateObj` below?

williamchong (changes requested):

src/routes/users/registerLogin.js
```diff
@@ -263,6 +263,10 @@ router.post(
         }
       });
 
+      if (!updateObj.displayName && !updateObj.locale) {
+        res.sendStatus(400);
+        return;
+      }
```
williamchong: Since we delete every key that is undefined in line 260, checking `Object.keys(updateObj).length` would allow us to check for invalid payload without hardcoding

williamchong (changes requested):

src/routes/users/registerLogin.js
WeiJun0827: Got it, I see

src/routes/users/registerLogin.js
```diff
@@ -263,6 +263,10 @@ router.post(
         }
       });
 
+      if (Object.keys(updateObj).length === 0) {
```
williamchong:
```suggestion
      if (!Object.keys(updateObj).length) {
```
Prefer truthy/falsey comparison unless we actually need to check the explicit value/type

src/routes/users/registerLogin.js
```diff
@@ -263,6 +263,10 @@ router.post(
...
+        res.sendStatus(400);
```
williamchong: Please add a message `INVALID_PAYLOAD`

williamchong (changes requested):

src/routes/users/registerLogin.js
```diff
@@ -263,7 +263,8 @@ router.post(
         }
       });
 
-      if (Object.keys(updateObj).length === 0) {
+      if (!Object.keys(updateObj).length) {
+        res.statusMessage = 'INVALID_PAYLOAD';
         res.sendStatus(400);
```
williamchong: In express 4.0 we can write
```suggestion
        res.status(400).send('INVALID_PAYLOAD');
```

williamchong (changes requested):

src/routes/users/registerLogin.js
WeiJun0827: Got it~! I was wondering if I should follow the coding style of the rest of code.

williamchong: We use `sendStatus` if we don't want to modify the message

src/routes/users/registerLogin.js
```diff
@@ -263,6 +263,10 @@ router.post(
         }
       });
 
+      if (!Object.keys(updateObj).length) {
+        res.status(400).send('INVALID_PAYLOAD');
+        return;
```
williamchong: I just realize we can actually
```suggestion
        throw new ValidationError('INVALID_PAYLOAD');
```
and let the error handler handle the http part
Sorry for missing this and please change to using ValidationError

williamchong (changes requested):

PR #226, opened 2021-03-16
✨ Split file upload and json api in user register/update

WeiJun0827:


src/routes/users/registerLogin.js
```diff
@@ -74,12 +77,26 @@ function csrfCheck(req, res, next) {
   }
 }
 
+// deprecated
+function isJson(req) {
+  return req.is('application/json') === 'application/json';
```
williamchong: `req.is` allow wildcard checking, if we are checking exact match `application/json` we might as well just check content-type

src/routes/users/registerLogin.js
```diff
@@ -74,12 +77,26 @@ function csrfCheck(req, res, next) {
...
+}
+
+// deprecated
+function loadRegisterMiddlewareByContentType(req, res, next) {
+  if (!isJson(req)) {
+    csrfCheck(req, res, () => {
+      bodyParser.urlencoded({ extended: false })(req, res, () => {
```
williamchong: Should check for input param `error` in all callback function, just in case

src/routes/users/registerLogin.js
```diff
@@ -99,7 +116,14 @@ router.post(
           } = req.body;
           if (!idToken) throw new ValidationError('ID_TOKEN_MISSING');
           if (!accessToken) throw new ValidationError('ACCESS_TOKEN_MISSING');
-          const authCoreUser = authCoreJwtVerify(idToken);
+          let authCoreUser;
+          try {
+            authCoreUser = authCoreJwtVerify(idToken);
+          } catch (err) {
+            console.error(err);
```
williamchong: No need to `console.error `since `ValidationError` is handled in error handler

src/routes/users/registerLogin.js
```diff
@@ -196,19 +220,29 @@ router.post(
   },
 );
 
+// deprecated
+function loadUpdateMiddlewareByContentType(req, res, next) {
```
williamchong: What is the different between this and the `loadRegisterMiddlewareByContentType` above?

src/routes/users/registerLogin.js
```diff
@@ -221,29 +255,27 @@ router.post(
...
       const updateObj = {
         displayName,
         isEmailEnabled,
-        avatar: avatarUrl,
         locale,
       };
+
+      // deprecated
+      let avatarUrl;
+      if (!isJson(req)) {
```
williamchong: Using a more verbose variable `isLegacyReq = !isJson(req)` is prefered

test/api/user.test.js
```diff
@@ -49,7 +50,7 @@ test.serial('USER: Login user. Case: success', async (t) => {
   t.is(res.status, 200);
 });
 
-test.serial('USER: Edit user. Case: success', async (t) => {
+test.serial('USER: Edit user by Json. Case: success', async (t) => {
```
williamchong:
```suggestion
test.serial('USER: Edit user by JSON. Case: success', async (t) => {
```

williamchong (changes requested):

src/routes/users/registerLogin.js
WeiJun0827: Would that means we prefer to use `req.get('Content-Type') === 'application/json'` here?

src/routes/users/registerLogin.js
WeiJun0827: `loadRegisterMiddlewareByContentType` applies `csrfCheck`, `bodyParser.urlencoded`, and `multer.single` functions.
`loadUpdateMiddlewareByContentType` applies `csrf`, `bodyParser.urlencoded`, and `multer.single` functions.
The difference is the csrf functions they refer.

src/routes/users/registerLogin.js
williamchong:
```suggestion
  return !!req.is('application/json');
```

williamchong (approved):

PR #227, opened 2021-03-19
🥅 Add JWT error handler

WeiJun0827:


src/routes/users/registerLogin.js
```diff
@@ -99,7 +99,13 @@ router.post(
...
           if (!idToken) throw new ValidationError('ID_TOKEN_MISSING');
           if (!accessToken) throw new ValidationError('ACCESS_TOKEN_MISSING');
-          const authCoreUser = authCoreJwtVerify(idToken);
+          let authCoreUser;
+          try {
+            authCoreUser = authCoreJwtVerify(idToken);
+          } catch (err) {
+            throw new ValidationError('ID_TOKEN_INVALID');
+          }
+
```
williamchong, line 109: Should check `authCoreUser` exists, throw error if not exists here?

williamchong (changes requested):

PR #239, opened 2021-04-30
✨ Support cosmos wallet signing register/login

williamchong:


src/util/api/users/index.js
```diff
@@ -157,6 +158,27 @@ export function checkSignPayload(from, payload, sign) {
...
+  signature, publicKey, message, cosmosWallet,
+}) {
+  const verified = verifyCosmosSignInPayload({
+    signature, publicKey, message, cosmosWallet,
+  });
+  if (!verified) {
+    throw new ValidationError('INVALID_SIGNATURE');
+  }
+  const { memo } = JSON.parse(message);
+  const actualPayload = JSON.parse(memo.substr(memo.indexOf('{')));
```
nwingt: Should wrap with try-catch for syntax error?

nnkken (approved):

PR #242, opened 2021-05-27
🔨 Add paymentRedirectWhiteList get/post field

WeiJun0827:


src/routes/users/preferences.js
```diff
@@ -52,6 +65,20 @@ router.post('/preferences', jwtAuth('write:preferences'), async (req, res, next)
       payload.creatorPitch = creatorPitch.slice(0, i);
     }
 
+    if (paymentRedirectWhiteList !== undefined) {
```
WeiJun0827: Just a minor question, would this be preferred?
```suggestion
    if (paymentRedirectWhiteList) {
```

williamchong: I think the issue is that the API user might want to just update other fields, say `locale`, without affecting `paymentRedirectWhiteList`. In this case, what would you expect the payload for the following 4 cases would be?
1. Remove everything in `paymentRedirectWhiteList`
2. Remove one entry in `paymentRedirectWhiteList`
3. Add one entry in `paymentRedirectWhiteList`
4. Do not change `paymentRedirectWhiteList`

WeiJun0827: May I assume the UI will get the prior settings first? If so, the field may remain previous  `paymentRedirectWhiteList` on UI for the user's further edit. And send the payload with:
1. `"paymentRedirectWhiteList": []`
2. `"paymentRedirectWhiteList": [ ... (one URL removed)]`
3. `"paymentRedirectWhiteList": [ ... (one URL added)]`
4. `"paymentRedirectWhiteList": [ ... (untouched)]`

williamchong: Since in 1. you assume empty array is delete, I think if `paymentRedirectWhiteList` is `undefined` we should leave it untouched. We should delete it if its null or empty array

WeiJun0827: I see, so the `!== undefined` part is necessary for our scenario. But I didn't expect the null case, should I add:
```
if (paymentRedirectWhiteList === null) {
    payload.paymentRedirectWhiteList = [];
}
```
at the beginning of the function right?
Or I could just simply see null as another not array payload and throw an exception?

williamchong: hmm i think we should add a special case and treat null as delete

WeiJun0827: No problem, code has been updated!

src/routes/users/preferences.js
```diff
@@ -52,6 +65,24 @@ router.post('/preferences', jwtAuth('write:preferences'), async (req, res, next)
...
+      if (paymentRedirectWhiteList === null) {
+        payload.paymentRedirectWhiteList = [];
+      } else {
+        if (!Array.isArray(paymentRedirectWhiteList)) {
+          res.status(400).send('INVALID_PAYMENT_REDIRECT_WHITELIST');
+          return;
+        }
+        for (let i = 0; i < paymentRedirectWhiteList.length; i += 1) {
+          if (!isValidHttpUrl(paymentRedirectWhiteList[i])) {
```
williamchong: should use `.some()`?

src/routes/users/preferences.js
```diff
@@ -52,6 +65,24 @@ router.post('/preferences', jwtAuth('write:preferences'), async (req, res, next)
       payload.creatorPitch = creatorPitch.slice(0, i);
     }
 
+    if (paymentRedirectWhiteList !== undefined) {
+      if (paymentRedirectWhiteList === null) {
+        payload.paymentRedirectWhiteList = [];
+      } else {
```
williamchong: The for loop below loops through `paymentRedirectWhiteList` as an array, should the null case be assigning to some variable, instead of assigning to payload? Also we can skip the else case if we assign to a variable then check it instead

williamchong (changes requested):

williamchong:
Please also add some simple test cases

src/routes/users/preferences.js
WeiJun0827: Cool, I will use .some() instead.
```
      if (paymentRedirectWhiteList.some(x => !isValidHttpUrl(x))) {
        res.status(400).send('INVALID_PAYMENT_REDIRECT_URL');
        return;
      }
```

test/api/user.test.js
```diff
@@ -566,3 +567,165 @@ test('USER: Check New User Info: Email Already exist', async (t) => {
...
+  const { paymentRedirectWhiteList: whitelist } = res.data;
+  t.is(res.status, 200);
+  t.is(whitelist.length, 3);
+  t.true(whitelist.includes('http://example1.com/'));
+  t.true(whitelist.includes('http://example2.com/'));
+  t.true(whitelist.includes('http://example3.com/'));
+});
+
+test('USER: Empty payment redirect whitelist with empty array. Case: Success', async (t) => {
+  const user = testingUser1;
```
williamchong: Testcases that reuse testingUser1 for post/get combo test that returns 200 should all be serial
otherwise potential race condition between cases

williamchong (changes requested):

test/api/user.test.js
WeiJun0827: Right... forget to change username at the last second

WeiJun0827: I think I would just use the same username `testingUser1` for all post & get cases, to make the code unified.

test/api/user-preferences.test.js
```diff
@@ -109,3 +109,165 @@ test('USER: Set user preferences (Creator pitch). Case: failed', async (t) => {
   });
   t.is(res.status, 400);
 });
+
+test('USER: Get user preferences. Case: Success', async (t) => {
```
williamchong: i think this case already exists in the beginning of this file. We should modify the first test instead of adding a new case  if you want to test for payment redirect url GET.

williamchong (changes requested):

williamchong (approved):

williamchong:
@nwingt please also see if you have any comment

nwingt (approved):
I am fine with the current implementation but do consider rename whiteList to whitelist for consistency

williamchong:
Maybe we can serve both version then remove one later, in another PR

PR #264, opened 2022-01-08
✨ Add ISCN route

williamchong:
- Add ISCN route `/new` which register metadata to ISCN
- Add ISCN route `/upload` which accept multipart files which uploads to Arweave then register metadata to ISCN
https://api.docs.like.co/#3312f20a-92f9-4af9-a492-c0f243304d60


src/util/cosmos/iscn.js
```diff
@@ -1,89 +1,39 @@
...
-  };
-  const signed = await signTransaction({
-    signer: cosmosSigner,
-    accNum,
-    stdTx,
-    sequence,
-  });
-  return signed;
+export async function getISCNSigningAddress() {
+  if (!signingWallet) return '';
```
WeiJun0827: Will it be a problem if this condition statement happen? Or actively initiate `signingWallet`?
```suggestion
  if (!signingWallet) getISCNSigningClient();
```

williamchong:
Updated and also refactor account number fetching

WeiJun0827 (approved):

PR #274, opened 2022-02-11
⚡️ Use media network cdn for arweave api

williamchong:
- Reduce anchor api call by sharing anchor for all tx
- Use media network cdn for non-gql calls

### DIFF
--- a/src/util/arweave.js
+++ b/src/util/arweave.js
@@ -20,8 +20,15 @@ const IPFS_CONSTRAINT = 'v0.1';
 
 const jwk = require('../../config/arweave-key.json');
 
-const arweave = Arweave.init({
+const arweaveGraphQL = Arweave.init({
   host: 'arweave.net',
+  port: 443,
+  protocol: 'https',
+  timeout: 60000,
+});
+
+const arweave = Arweave.init({
+  host: 'gateway.arweave.network',
   port: 443,
   protocol: 'https',
   timeout: 60000,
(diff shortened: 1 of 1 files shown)


src/util/arweave.js
```diff
@@ -288,32 +301,33 @@ export async function uploadFilesToArweave(files, arweaveIdList, checkDuplicate
...
     let arweaveId = arweaveIds[i];
     let ipfsHash = ipfsHashes[i];
     if (!ipfsHash) {
       ipfsHash = await getFileIPFSHash(f);
     }
     if (!arweaveId) {
       [arweaveId] = await Promise.all([
-        submitToArweave(f, ipfsHash, checkDuplicate),
+        submitToArweave(f, ipfsHash, {
+          checkDuplicate,
```
WeiJun0827: Does `submitToArweave()` accept `{ checkDuplicate }` pareameter?

WeiJun0827 (approved):

src/util/arweave.js
```diff
@@ -122,7 +129,10 @@ export async function estimateARPrice(data, checkDuplicate = true) {
       };
     }
   }
-  const transaction = await arweave.createTransaction({ data: buffer }, jwk);
+  const transaction = await arweave.createTransaction({
+    data: buffer,
+    last_tx: 'stub_for_estimate',
```
WeiJun0827, line 134: Curious about the function of adding this config?

williamchong: if `last_tx` is not set, arweavejs will query `/tx_anchor` API, which cost ~300ms-3s and has no meaning for estimation

WeiJun0827: Got it, thanks!

PR #279, opened 2022-04-19
🔨 Check public key in `userId` rather than cosmos address

WeiJun0827:


src/routes/users/registerLogin.js
```diff
@@ -162,7 +162,7 @@ router.post(
           payload.displayName = user;
           payload.email = email;
           payload.isEmailVerified = false;
-          platformUserId = cosmosWallet;
+          platformUserId = publicKey;
```
nwingt, line 175: the above `cosmosWallet` should be generated by `publicKey` instead?

WeiJun0827: I see. So the frontend no longer provides `cosmosWallet`.

nwingt: Is it also needed to set `likeWallet` as well?

WeiJun0827: Hmm, right, so that we have a `likeWallet` field in DB.

williamchong (commented):
Should we change `checkCosmosSignPayload` to check for public key instead of cosmosWallet in memo?

williamchong:
Due to pubkey not migrate-able, we would be doing likeWallet as address first
https://github.com/likecoin/likecoin-api-public/pull/281

PR #281, opened 2022-04-21
👽️ Add likeWallet support

williamchong:
- Allow checking signature via like or cosmos prefix
- Add likeWallet and cosmosWallet field for user
- Add likeWallet and cosmosWallet field in auth as long as a wallet exists, no matter platform

### DIFF
--- a/src/routes/users/platforms/index.js
+++ b/src/routes/users/platforms/index.js
@@ -13,6 +13,7 @@ import { ValidationError } from '../../../util/ValidationError';
 import { jwtAuth } from '../../../middleware/jwt';
 import publisher from '../../../util/gcloudPub';
 import { authCoreJwtSignToken, authCoreJwtVerify } from '../../../util/jwt';
+import { convertAddressPrefix } from '../../../util/cosmos';
 
 const router = Router();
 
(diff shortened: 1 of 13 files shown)


src/routes/users/platforms/index.js
```diff
@@ -140,9 +145,18 @@ router.post('/login/:platform/add', jwtAuth('write'), async (req, res, next) =>
             }
           });
         }
+        if (likeWalletQuery.docs.length > 0) {
+          likeWalletQuery.forEach((doc) => {
+            const docUser = doc.id;
+            if (user !== docUser) {
+              throw new ValidationError('AUTHCORUE_WALLET_DUPLICATED');
```
nwingt:
```suggestion
              throw new ValidationError('AUTHCORE_WALLET_DUPLICATED');
```

src/util/api/users/register.js
```diff
@@ -107,13 +108,17 @@ export async function handleUserRegistration({
   }
 
   if (!checkUserNameValid(user)) throw new ValidationError('Invalid user name');
-  if (!checkCosmosAddressValid(cosmosWallet)) {
+  if (!checkCosmosAddressValid(cosmosWallet, 'cosmos')) {
+    throw new ValidationError('invalid cosmos wallet');
+  }
+  if (!checkCosmosAddressValid(likeWallet, 'like')) {
```
nwingt, line 114: The error should be 'invalid like wallet'?

nwingt (approved):

PR #283, opened 2022-04-22
Update for avatar upload & fix breaking change on noise-js

nwingt:


WeiJun0827 (approved):

williamchong (approved):

src/util/fileupload.js
```diff
@@ -58,7 +58,7 @@ export async function handleAvatarUploadAndGetURL(user, file, avatarSHA256) {
     filename: `likecoin_store_user_${user}_${IS_TESTNET ? 'test' : 'main'}`,
     mimetype: file.mimetype,
   });
-  return avatarUrl;
+  return `${avatarUrl}&${hash256.substring(0, 7)}`;
```
williamchong: I think md5 is better in this case, since you only take 7 char substring and md5 is shorter than sha256

williamchong (approved):
from which lib this `md5-hex` is from though?

nwingt:
> from which lib this `md5-hex` is from though?

@williamchong  ava

williamchong:
`ava` is in `devDependencies`, that means we do need to add `md5-hex` to `dependencies`

PR #297, opened 2022-06-20
✨ Allow querying likerid via wallet

williamchong:


src/util/api/users/getPublicInfo.js
```diff
@@ -45,7 +40,17 @@ export async function getUserWithCivicLikerProperties(id) {
       payload.isExpiredCivicLiker = true;
     }
   }
+  return payload;
+}
+
+export async function getUserWithCivicLikerProperties(id) {
+  const [userDoc] = await Promise.all([
+    dbRef.doc(id).get(),
+  ]);
```
nwingt: Why use Promise.all here?

williamchong: Removed

PR #300, opened 2022-07-07
✨ GET /id min add description

Tsai-Hsueh-Kuan:
✨ GET /id min add description


src/util/ValidationHelper.js
```diff
@@ -110,6 +111,7 @@ export function filterUserDataMin(userObject, types = []) {
     isCivicLikerTrial,
     isSubscribedCivicLiker,
     civicLikerSince,
+    description: description || '',
```
williamchong: can we just do `description`?
```suggestion
    description,
```

williamchong (changes requested):

williamchong (approved):

PR #302, opened 2022-07-13
🔨 Recover querying user info by address endpoint

WeiJun0827:


williamchong (approved):
Can you help me to add a testcase for this API? Thanks

PR #304, opened 2022-07-19
♻️ Split signing rpc into separated config

williamchong:
we want query to be loadbalance/cached, but signing to be a same node for increasing sequence

### DIFF
--- a/src/util/cosmos/iscn.js
+++ b/src/util/cosmos/iscn.js
@@ -3,9 +3,7 @@ import { DirectSecp256k1Wallet } from '@cosmjs/proto-signing';
 import { ISCNQueryClient, ISCNSigningClient } from '@likecoin/iscn-js';
 import { getAccountInfo } from '.';
 import { COSMOS_PRIVATE_KEY } from '../../../config/secret';
-import {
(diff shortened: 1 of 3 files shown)


src/util/cosmos/iscn.js
```diff
@@ -3,9 +3,9 @@ import { DirectSecp256k1Wallet } from '@cosmjs/proto-signing';
 import { ISCNQueryClient, ISCNSigningClient } from '@likecoin/iscn-js';
 import { getAccountInfo } from '.';
 import { COSMOS_PRIVATE_KEY } from '../../../config/secret';
-import {
-  COSMOS_RPC_ENDPOINT,
-} from '../../../config/config';
+import { COSMOS_RPC_ENDPOINT, COSMOS_SIGNING_RPC_ENDPOINT } from '../../../config/config';
+
+export { parseTxInfoFromIndexedTx } from '@likecoin/iscn-js/dist/messages/parsing';
```
nwingt: What is this export for?

williamchong: conflict.. will remove

PR #308, opened 2022-07-27
Feature/social graph API

lancatlin:
+ add `/likernft/collector?creator=` and `/likernft/creator?collector=` API
+ get original response from indexer and aggregate price from firestore
+ order by total value desc

### DIFF
--- a/src/routes/likernft/socialgraph.js
+++ b/src/routes/likernft/socialgraph.js
@@ -6,7 +6,7 @@ const router = Router();
 
 router.get('/collector', async (_, res) => {
   const result = await getCollector(
-    'like156gedr03g3ggwktzhygfusax4df46k8dh6w0me',
+    'like1qv66yzpgg9f8w46zj7gkuk9wd2nrpqmca3huxf',
   );
   res.send(result);
 });
--- a/src/routes/likernft/socialgraph.js
+++ b/src/routes/likernft/socialgraph.js
@@ -1,13 +1,26 @@
 import { Router } from 'express';
 
-import { getCollector } from '../../util/api/likernft/socialgraph';
+import { getCollector, getCreator } from '../../util/api/likernft/socialgraph';
 
 const router = Router();
 
-router.get('/collector', async (_, res) => {
-  const result = await getCollector(
-    'like1qv66yzpgg9f8w46zj7gkuk9wd2nrpqmca3huxf',
-  );
+router.get('/collector', async (req, res) => {
+  const { creator } = req.query;
+  if (!creator) {
+    res.status(401).send({ error: 'creator is required' });
+    return;
+  }
+  const result = await getCollector(creator);
+  res.send(result);
+});
+
+router.get('/creator', async (req, res) => {
+  const { collector } = req.query;
+  if (!collector) {
+    res.status(401).send({ error: 'creator is required' });
+    return;
+  }
+  const result = await getCreator(collector);
   res.send(result);
 });
 
--- a/src/routes/likernft/socialgraph.js
+++ b/src/routes/likernft/socialgraph.js
@@ -1,0 +1,14 @@
+import { Router } from 'express';
+
+import { getCollector } from '../../util/api/likernft/socialgraph';
+
+const router = Router();
+
+router.get('/collector', async (_, res) => {
+  const result = await getCollector(
+    'like156gedr03g3ggwktzhygfusax4df46k8dh6w0me',
+  );
+  res.send(result);
+});
+
+export default router;
--- a/src/routes/likernft/socialgraph.js
+++ b/src/routes/likernft/socialgraph.js
@@ -4,24 +4,32 @@ import { getCollector, getCreator } from '../../util/api/likernft/socialgraph';
 
 const router = Router();
 
-router.get('/collector', async (req, res) => {
-  const { creator } = req.query;
-  if (!creator) {
-    res.status(401).send({ error: 'creator is required' });
-    return;
+router.get('/collector', async (req, res, next) => {
+  try {
+    const { creator } = req.query;
+    if (!creator) {
+      res.status(400).json({ error: 'MISSING_CREATOR' });
+      return;
+    }
+    const result = await getCollector(creator);
+    res.json(result);
+  } catch (err) {
+    next(err);
   }
-  const result = await getCollector(creator);
-  res.send(result);
 });
 
-router.get('/creator', async (req, res) => {
-  const { collector } = req.query;
-  if (!collector) {
-    res.status(401).send({ error: 'creator is required' });
-    return;
+router.get('/creator', async (req, res, next) => {
+  try {
+    const { collector } = req.query;
+    if (!collector) {
+      res.status(400).json({ error: 'MISSING_COLLECTOR' });
+      return;
+    }
+    const result = await getCreator(collector);
+    res.json(result);
+  } catch (err) {
+    next(err);
   }
-  const result = await getCreator(collector);
-  res.send(result);
 });
 
 export default router;
--- a/src/util/api/likernft/socialgraph.js
+++ b/src/util/api/likernft/socialgraph.js
@@ -61,7 +61,6 @@ async function getCollector(creator) {
       collectors: await aggregate(res.data.collectors),
     };
 
-    console.dir(payload, { depth: null });
     return payload;
   } catch (err) {
     console.log(err.message);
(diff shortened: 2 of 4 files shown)


src/routes/likernft/socialgraph.js
```diff
@@ -0,0 +1,27 @@
+import { Router } from 'express';
+
+import { getCollector, getCreator } from '../../util/api/likernft/socialgraph';
+
+const router = Router();
+
+router.get('/collector', async (req, res) => {
+  const { creator } = req.query;
+  if (!creator) {
+    res.status(401).send({ error: 'creator is required' });
```
williamchong: This seems to be a `400` error instead of `401`
also we use plaintext error e.g. `MISISNG_CREATOR`

src/routes/likernft/socialgraph.js
```diff
@@ -0,0 +1,27 @@
...
+    return;
+  }
+  const result = await getCollector(creator);
+  res.send(result);
+});
+
+router.get('/creator', async (req, res) => {
+  const { collector } = req.query;
+  if (!collector) {
+    res.status(401).send({ error: 'creator is required' });
```
williamchong: This seems to be a `400` error instead of `401`
also we use plaintext error e.g. `MISISNG_CREATOR`

src/routes/likernft/socialgraph.js
```diff
@@ -0,0 +1,27 @@
...
+const router = Router();
+
+router.get('/collector', async (req, res) => {
+  const { creator } = req.query;
+  if (!creator) {
+    res.status(401).send({ error: 'creator is required' });
+    return;
+  }
+  const result = await getCollector(creator);
+  res.send(result);
```
williamchong: Prefer `json()` if the result is json

src/routes/likernft/socialgraph.js
```diff
@@ -0,0 +1,27 @@
...
+});
+
+router.get('/creator', async (req, res) => {
+  const { collector } = req.query;
+  if (!collector) {
+    res.status(401).send({ error: 'creator is required' });
+    return;
+  }
+  const result = await getCreator(collector);
+  res.send(result);
```
williamchong: Prefer `json()` if the result is json

src/routes/likernft/socialgraph.js
```diff
@@ -0,0 +1,27 @@
...
+});
+
+router.get('/creator', async (req, res) => {
+  const { collector } = req.query;
+  if (!collector) {
+    res.status(401).send({ error: 'creator is required' });
+    return;
+  }
+  const result = await getCreator(collector);
```
williamchong: Please wrap the whole route in `try catch` where in the catch block we call `next(error)`

src/routes/likernft/socialgraph.js
```diff
@@ -0,0 +1,27 @@
...
+
+const router = Router();
+
+router.get('/collector', async (req, res) => {
+  const { creator } = req.query;
+  if (!creator) {
+    res.status(401).send({ error: 'creator is required' });
+    return;
+  }
+  const result = await getCollector(creator);
```
williamchong: Please wrap the whole route in `try catch` where in the catch block we call `next(error)`

src/util/api/likernft/socialgraph.js
```diff
@@ -0,0 +1,88 @@
...
+      `/likechain/likenft/v1/collector?creator=${creator}`,
+    );
+    const payload = {
+      ...res.data,
+      collectors: await aggregate(res.data.collectors),
+    };
+
+    return payload;
+  } catch (err) {
+    console.log(err.message);
```
williamchong: Please use `console.error` for actual error.
Any reason we don't throw this error to let the route handle instead of returning empty?

src/util/api/likernft/socialgraph.js
```diff
@@ -0,0 +1,88 @@
...
+    );
+    const payload = {
+      ...res.data,
+      creators: await aggregate(res.data.creators),
+    };
+
+    return payload;
+  } catch (err) {
+    console.log(err.message);
+    return { error: err.message };
```
williamchong: Please use `console.error` for actual error.
I think the issue of returning `error` object is that we don't check if the return value contains `error` and throw http 4xx/5xx accordingly
For now I think just throwing `err` is simpler

src/util/api/likernft/socialgraph.js
```diff
@@ -0,0 +1,88 @@
...
+  const promises = [];
+  const newAccounts = [];
+  Object.entries(accounts).forEach(([owner, value]) => {
+    const account = {
+      account: owner,
+      collections: [],
+      count: value.count,
+      totalValue: 0,
+    };
+    value.collections.forEach(
```
williamchong, line 17: What is the length of `value.collections` are we expecting? depends on the length we might want `getLatestNFTPriceAndInfo` to be blocking or parallel

src/util/api/likernft/socialgraph.js
```diff
@@ -0,0 +1,88 @@
...
+                  classId,
+                  count,
+                  price,
+                  totalValue: count * price,
+                });
+                account.totalValue += count * price;
+                resolve(true);
+              })
+              .catch((err) => {
+                console.log(err, iscnPrefix, classId);
```
williamchong: What kind of error are we expecting? Also please use `console.error` for actual error

src/util/api/likernft/socialgraph.js
```diff
@@ -0,0 +1,88 @@
...
+      collections: [],
+      count: value.count,
+      totalValue: 0,
+    };
+    value.collections.forEach(
+      ({ iscn_id_prefix: iscnPrefix, class_id: classId, count }) => {
+        promises.push(
+          new Promise((resolve) => {
+            getLatestNFTPriceAndInfo(iscnPrefix, classId)
+              .then(({ price }) => {
```
williamchong: if the NFT is soldout, price might be undefined, we should use `lastSoldPrice` in this case

src/util/api/likernft/socialgraph.js
```diff
@@ -0,0 +1,88 @@
...
+      account: owner,
+      collections: [],
+      count: value.count,
+      totalValue: 0,
+    };
+    value.collections.forEach(
+      ({ iscn_id_prefix: iscnPrefix, class_id: classId, count }) => {
+        promises.push(
+          new Promise((resolve) => {
+            getLatestNFTPriceAndInfo(iscnPrefix, classId)
```
williamchong: I think one way we can optimize this API is to add a flag in `getLatestNFTPriceAndInfo`. We would want to ignore `getLowerestUnsoldNFT` and just call `likeNFTCollection.doc(iscnPrefix).get()` which is much less expensive to call

src/util/api/likernft/socialgraph.js
```diff
@@ -0,0 +1,88 @@
...
+    const account = {
+      account: owner,
+      collections: [],
+      count: value.count,
+      totalValue: 0,
+    };
+    value.collections.forEach(
+      ({ iscn_id_prefix: iscnPrefix, class_id: classId, count }) => {
+        promises.push(
+          new Promise((resolve) => {
```
williamchong: Just push `getLatestNFTPriceAndInfo.then()` which is already a promise, no need to `new Promise`

src/util/api/likernft/socialgraph.js
```diff
@@ -0,0 +1,88 @@
...
+              })
+              .catch((err) => {
+                console.log(err, iscnPrefix, classId);
+                account.collections.push({
+                  iscnPrefix,
+                  classId,
+                  count,
+                  price: 0,
+                  totalValue: 0,
+                });
```
williamchong: I think we can have a default estimation of total price according to current pricing table?
(In fact I am not sure if we can skip the firestore part entirely, and just query by lookup table given we never change the price)

williamchong (changes requested):

src/util/api/likernft/socialgraph.js
lancatlin: Some NFT may be issued by user themselves so we cannot find record in firestore

src/util/api/likernft/socialgraph.js
lancatlin: `value.collections` is related to how many works the creator has published. I think it would be less than 100 for normal situation? Because most of collections in different collector may be duplicated, I think doing some cache will help.

PR #313, opened 2022-07-29
✨ Add transferring NFT endpoint

WeiJun0827:


src/util/api/likernft/transfer.js
```diff
@@ -0,0 +1,28 @@
...
+  txTimestamp,
+}) {
+  const iscnPrefix = getISCNPrefixDocName(iscnId);
+  const iscnRef = likeNFTCollection.doc(iscnPrefix);
+  const classRef = iscnRef.collection('class').doc(classId);
+  const nftRef = classRef.collection('nft').doc(nftId);
+  await db.runTransaction(async (t) => {
+    const nftDoc = await t.get(nftRef);
+    if (!nftDoc.exists) throw new ValidationError('NFT_NOT_FOUND');
+    const { lastUpdateTimestamp: dbTimestamp } = nftDoc.data();
```
williamchong: how about `lastUpdateTimestamp: dbTimestamp = 0`

src/routes/likernft/transfer.js
```diff
@@ -0,0 +1,49 @@
...
+
+const router = Router();
+
+router.post(
+  '/transfer',
+  async (req, res, next) => {
+    try {
+      const { tx_hash: txHash } = req.query;
+      if (!txHash) throw new ValidationError('MISSING_TX_HASH');
+      const { data } = await axios.get(`${COSMOS_LCD_INDEXER_ENDPOINT}/cosmos/tx/v1beta1/txs/${txHash}`);
```
williamchong: Would just getting via stargateclient's getTx() be easier? this path does not hit indexer db anyway

src/routes/likernft/transfer.js
```diff
@@ -0,0 +1,49 @@
...
+      const {
+        class_id: classId,
+        id: nftId,
+        sender,
+        receiver,
+      } = data.tx.body.messages[0];
```
williamchong: Should we find target event in message or it is always first? what if there does not exists class_id in first message? i.e random message hash

src/routes/likernft/transfer.js
williamchong: Also what if tx not found

src/routes/likernft/transfer.js
```diff
@@ -0,0 +1,49 @@
...
+      });
+      res.json({
+        txTimestamp,
+        txHash,
+        iscnId,
+        classId,
+        nftId,
+        fromAddress: sender,
+        toAddress: receiver,
+      });
```
williamchong, line 39: No response?

williamchong (changes requested):

src/util/api/likernft/transfer.js
WeiJun0827: great!

src/routes/likernft/transfer.js
WeiJun0827: Do return `res.json()` if it's successful. Is it OK?

williamchong: Oh I missed the `res.json()` part, nevermind

williamchong (approved):

PR #316, opened 2022-07-31
♻️ Refactor get iscn prefix function to throw error

williamchong:


src/util/cosmos/iscn.js
```diff
@@ -54,8 +54,8 @@ export async function getISCNSigningAddressInfo() {
 }
 
 export function getISCNPrefix(input) {
-  const res = /^(iscn:\/\/likecoin-chain\/[A-Za-z0-9-_]+)\/[0-9]+$/.exec(input);
-  if (!res) return input;
+  const res = /^(iscn:\/\/likecoin-chain\/[A-Za-z0-9-_]+)(?:\/?([0-9]*))?$/.exec(input);
```
nnkken:
```suggestion
  const res = /^(iscn:\/\/likecoin-chain\/[A-Za-z0-9-_]+)(?:\/([0-9]+))?$/.exec(input);
```

Reference: the ISCN ID regex pattern in LikeCoin chain source code:
```Go
const IscnIdRegexpPattern = "iscn://([-_.:=+,a-zA-Z0-9]+)/([-_.:=+,a-zA-Z0-9]+)(?:/([0-9]+))?$"
```

src/util/cosmos/iscn.js
```diff
@@ -54,8 +54,8 @@ export async function getISCNSigningAddressInfo() {
 }
 
 export function getISCNPrefix(input) {
-  const res = /^(iscn:\/\/likecoin-chain\/[A-Za-z0-9-_]+)\/[0-9]+$/.exec(input);
-  if (!res) return input;
+  const res = /^(iscn:\/\/likecoin-chain\/[A-Za-z0-9-_]+)(?:\/?([0-9]+))?$/.exec(input);
```
nnkken: The `/` is necessary for the group. The group (`(?:\/?([0-9]+))?`) either exist with `/`, or the whole group does not exist.
Current pattern will work because `[A-Za-z0-9-_]+` is greedy and will eat all numbers at the end, leaving nothing for the group. But it's confusing and not robust.

```suggestion
  const res = /^(iscn:\/\/likecoin-chain\/[A-Za-z0-9-_]+)(?:\/([0-9]+))?$/.exec(input);
```

williamchong: hmm what if I want to support tailing slash like `iscn://likecoin-chain/laa5PLHfQO2eIfiPB2-ZnFLQrmSXOgL-NvoxyBTXHvY1/`?

williamchong: `/^(iscn:\/\/likecoin-chain\/[A-Za-z0-9-_]+)(?:\/([0-9]*))?$/`?

WeiJun0827 (approved):

PR #325, opened 2022-08-06
✨ Add check user ts in purchase

Tsai-Hsueh-Kuan:
✨ Add check user ts in purchase

### DIFF
--- a/src/routes/likernft/purchase.js
+++ b/src/routes/likernft/purchase.js
@@ -12,6 +12,8 @@ import { fetchISCNIdAndClassId } from '../../middleware/likernft';
 import publisher from '../../util/gcloudPub';
 import { PUBSUB_TOPIC_MISC } from '../../constant';
 
+const API_EXPIRATION_BUFFER_TIME = 5000;
+
 const router = Router();
 
 router.get(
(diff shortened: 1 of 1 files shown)


src/routes/likernft/purchase.js
```diff
@@ -47,7 +49,8 @@ router.post(
   fetchISCNIdAndClassId,
   async (req, res, next) => {
     try {
-      const { tx_hash: txHash } = req.query;
+      const { tx_hash: txHash, user_ts: userTs } = req.query;
+      if (Date.now() - userTs > API_EXPIRATION_BUFFER_TIME) throw new ValidationError('USER_TS_ERROR');
```
williamchong: USER_TIME_OUT_SYNC

williamchong (changes requested):

src/routes/likernft/purchase.js
```diff
@@ -47,7 +49,8 @@ router.post(
   fetchISCNIdAndClassId,
   async (req, res, next) => {
     try {
-      const { tx_hash: txHash } = req.query;
+      const { tx_hash: txHash, user_ts: userTs } = req.query;
+      if (Date.now() - userTs > API_EXPIRATION_BUFFER_TIME) throw new ValidationError('USER_TIME_OUT_SYNC');
```
williamchong: i think user_ts can be simply rename to ts
and we can skip checking if ts was not sent

williamchong (approved):

PR #334, opened 2022-08-16
Support ISCN ID and NFT class in oEmbed

nnkken:
Also fixed (or broken?) something according to the [oEmbed spec](https://oembed.com/):

 - `url` field in resposne is only for `photo` type, changed to `author_url` for user's Like Button
 - added `author_name` for user's Like Button

Also imported new `eslintrc.js` and made all `console.*` calls errors (need to explicitly disabling eslint `no-console` rule).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,22 @@
+module.exports = {
+  extends: 'airbnb-base',
+  rules: {
+    'no-console': 'error',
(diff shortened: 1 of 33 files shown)


williamchong (approved):

williamchong:
Add some test case for the iscn and nft case maybe?

PR #339, opened 2022-08-29
⚡️ Port arweave api performance patch from app-like-co

williamchong:


nwingt (approved):

Tsai-Hsueh-Kuan (approved):
The exact same code appears in applikeco, shouldl it be integrated later when we have time?

williamchong:
> The exact same code appears in applikeco, shouldl it be integrated later when we have time?

I think applikeco should call api.like.co in the future

PR #341, opened 2022-08-31
Fix cannot parse oEmbed requests

nnkken:
Now should support ISCN ID and NFT class by simply inputting `button.like.co/(ISCN ID or NFT class)`.
Also added `button.like.co/user/(user ID)` for some extreme(?) username case


williamchong:
I think the `/in/like` happens due to redirect here, maybe we should change this later..
https://github.com/likecoin/likecoin-button/blob/master/src/pages/_id/index.vue#L56

williamchong (approved):

PR #354, opened 2022-09-27
✨ Handle ISCN ownership change for `/sell`

WeiJun0827:


src/routes/likernft/user.js
```diff
@@ -12,20 +16,27 @@ router.get(
...
           .where('soldCount', '>', 0).get(),
-        // TODO: what if iscn owner changed?
-        likeNFTCollection.where('creatorWallet', '==', wallet).get(),
+        likeNFTCollection.where('ownerWallet', '==', wallet).get(),
       ]);
       const classIdSet = new Set();
+      const now = Date.now();
+      if (!updateCooldownMap[wallet] || now - updateCooldownMap[wallet] > UPDATE_COOLDOWN) {
+        updateCooldownMap[wallet] = now;
+        const ownedClassIds = await filterOwnedClassIds(ownedClassesQuery.docs, wallet);
```
williamchong, line 29: A performance hack would be skipping class Ids that exists in `sellingNftsQuery` when checking

src/util/api/likernft/user.js
```diff
@@ -0,0 +1,23 @@
...
+  const promises = iscnDocs.map(async (doc) => {
+    const iscnPrefix = decodeURIComponent(doc.id);
+    const owner = await getNFTISCNOwner(iscnPrefix);
+    if (owner && owner !== wallet) {
+      classIdSet.delete(doc.data().classId);
+      batch.update({ ownerWallet: owner });
+    }
+  });
+  await Promise.all(promises);
+  await batch.commit();
```
williamchong: What if the batch is empty?

src/routes/likernft/user.js
WeiJun0827: An edge case is that: 

> User A is selling a number of NFTs minted from one of his ISCN, and he changed the ownership of this ISCN to User B later. 

If we skip the class Ids that exist in `sellingNftsQuery` when checking, will we be able to update the ownership of this ISCN when calling `/user/walletA/sell` before these NFTs sold out?

williamchong: makes sense, keep the current change then

src/util/api/likernft/user.js
WeiJun0827: Fixed

williamchong: Another issue would be what if `iscnDocs` > max batch size

src/util/api/likernft/user.js
```diff
@@ -6,19 +6,23 @@ export async function filterOwnedClassIds(iscnDocs, wallet) {
...
   const promises = iscnDocs.map(async (doc) => {
     const iscnPrefix = decodeURIComponent(doc.id);
     const owner = await getNFTISCNOwner(iscnPrefix);
     if (owner && owner !== wallet) {
       classIdSet.delete(doc.data().classId);
       batch.update({ ownerWallet: owner });
-      isEmptyBatch = false;
+      count += 1;
+      if (count % 200 === 0) {
+        await batch.commit();
```
williamchong: what if count 200 is stuck at `await batch.commit();` then count 201 tries to call `batch.update` ? Seems like a race condition in `.map()` for me

williamchong: @nnkken please comment if you think there is better way to write this

WeiJun0827: Modified, not sure if we can do it like the below commit.

nnkken (approved):

PR #358, opened 2022-09-27
✨ Send all token to stakeholder for first sale

williamchong:


src/util/api/likernft/purchase.js
```diff
@@ -188,7 +186,10 @@ export async function handleNFTPurchaseTransaction({
   buyerWallet,
   granterWallet,
   feeWallet,
+  isResell,
 }, req) {
+  const STAKEHOLDERS_RATIO = isResell ? 0.1 : 1 - FEE_RATIO;
+  const SELLER_RATIO = isResell ? 1 - FEE_RATIO - STAKEHOLDERS_RATIO : 0;
```
nnkken: Seems no need to split 2 cases for `SELLER_RATIO`, simply use `1 - FEE_RATIO - STAKEHOLDERS_RATIO` as it will be computed as `0` when `!isResell`

nnkken (approved):

PR #360, opened 2022-10-05
✨ Return `iscnRecordTimestamp` for sorting created classIds in liker.land

WeiJun0827:


src/routes/likernft/metadata.js
```diff
@@ -31,6 +31,7 @@ router.get(
         iscnId: iscnPrefix,
         iscnOwner,
         iscnStakeholders: iscnData.stakeholders,
+        iscnTimestamp: iscnData.recordTimestamp,
```
williamchong: Not sure if `iscnRecordTimestamp` is better name

williamchong (approved):

PR #362, opened 2022-10-06
♻️ Update test LIKE

Tsai-Hsueh-Kuan:
♻️ Update test LIKE


williamchong:
Merged with https://github.com/likecoin/likecoin-api-public/pull/363

PR #366, opened 2022-10-07
🐛 Replace toFixed(0) with toString()

WeiJun0827:


williamchong:
The issue is `totalLIKE` in `getStakeholderMapFromParsedIscnData` is expected to be number, but its a string converted from bignumber currently

williamchong:
Close with https://github.com/likecoin/likecoin-api-public/pull/367

PR #379, opened 2022-10-24
✨ Add user delete api

williamchong:
What will be deleted:
- user data
- user auth data
- user subscription data + superlike status
- likecoin button url and iscn info related to liker ID

What will be kept
- general tx history
- civic liker like tx history

LikerId is designed not to be reuse-able after deletion due to tx record being kept


src/util/api/users/delete.js
```diff
@@ -0,0 +1,64 @@
...
+  const query = await likeButtonUrlCollection
+    .where('user', '==', user)
+    .get();
+  if (!query.docs.length) return;
+  let batch = db.batch();
+  let i;
+  for (i = 0; i < query.docs.length; i += 1) {
+    batch.delete(query.docs[i].ref);
+    if (i % BATCH_SIZE === 0) await batch.commit();
+    batch = db.batch();
```
nnkken: These 2 lines need to be in the same `if` block?

nnkken: Same for similar code below

test/api/user-delete.test.js
```diff
@@ -0,0 +1,187 @@
...
+  cosmosPrivateKeyDelete,
+} from './data';
+import axiosist from './axiosist';
+import {
+  signWithPrivateKey as signWithCosmos,
+} from './cosmos';
+
+const { jwtSign } = require('./jwt');
+
+test('USER: Delete authcore user. Case: fail, no authcore tokem', async (t) => {
```
nwingt:
```suggestion
test('USER: Delete authcore user. Case: fail, no authcore token', async (t) => {
```

test/api/user-delete.test.js
```diff
@@ -0,0 +1,187 @@
...
+  }, {
+    headers: {
+      Accept: 'application/json',
+      Cookie: `likecoin_auth=${token}`,
+    },
+  });
+  t.is(res.status, 400);
+});
+
+test('USER: Delete wallet user. Case: sucess', async (t) => {
```
nwingt:
```suggestion
test('USER: Delete wallet user. Case: success, async (t) => {
```

nwingt (changes requested):

nnkken (approved):

PR #382, opened 2022-10-27
✨ Support dynamic Liker avatar size

WeiJun0827:


src/routes/users/getPublicInfo.js
```diff
@@ -39,6 +43,37 @@ router.get('/id/:id/min', async (req, res, next) => {
...
 });
 
+
+router.get('/id/:id/avatar', async (req, res, next) => {
+  try {
+    const { id } = req.params;
+    const { size = '400' } = req.query;
+    const sizeNum = parseInt(size, 10);
+    if (Number.isNaN(sizeNum) || sizeNum <= 0 || sizeNum > 400) {
+      throw new ValidationError('Invalid size');
```
williamchong, line 44: I think we can cast the number in range using Math.min and Math.max instead of error

src/routes/users/getPublicInfo.js
```diff
@@ -39,6 +43,37 @@ router.get('/id/:id/min', async (req, res, next) => {
...
+    }
+
+    const avatar = await getUserAvatar(id);
+    if (avatar) {
+      const { data: stream } = await axios.get(avatar, {
+        responseType: 'stream',
+      });
+      const resizer = sharp()
+        .resize(sizeNum, sizeNum)
+        .jpeg();
```
williamchong: Use `content-type` header from stream to determine content type, don't hardcode jpeg

williamchong: In fact just setting `content-type` in `res` without handling any filetype logic in sharp might already do

williamchong (changes requested):

src/util/api/users/getPublicInfo.js
```diff
@@ -1,19 +1,37 @@
...
+function formatUserCivicLikerProperies(userDoc) {
+  const { id } = userDoc;
+  const data = userDoc.data();
+  const { civicLiker } = data;
   const payload = data;
   payload.user = id;
-  if (!avatar) {
-    payload.avatar = AVATAR_DEFAULT_PATH;
-  }
+  payload.avatar = `https://${API_EXTERNAL_HOSTNAME}/users/id/${id}/avatar`;
```
williamchong: Can we add `size` qs with default value here to hint there is a size configurable?
define `const DEFAULT_AVATAR_SIZE = 400` in `constant and
```suggestion
  payload.avatar = `https://${API_EXTERNAL_HOSTNAME}/users/id/${id}/avatar?size=${DEFAULT_AVATAR_SIZE}`;
```

src/routes/users/getPublicInfo.js
```diff
@@ -39,6 +34,35 @@ router.get('/id/:id/min', async (req, res, next) => {
   }
 });
 
+
+router.get('/id/:id/avatar', async (req, res, next) => {
+  try {
+    const { id } = req.params;
+    const { size: inputSizeStr = '400' } = req.query;
```
williamchong: Please refactor to `constant`

src/routes/users/getPublicInfo.js
```diff
@@ -39,6 +34,35 @@ router.get('/id/:id/min', async (req, res, next) => {
...
+      const { headers, data } = await axios.get(avatar, {
+        responseType: 'stream',
+      });
+      const resizer = sharp().resize(size, size);
+      const cacheTime = 3600;
+      res.set('Cache-Control', `public, max-age=${cacheTime}, s-maxage=${cacheTime}, stale-if-error=${ONE_DAY_IN_S}`);
+      res.type(headers['content-type'] || 'image/jpeg');
+      data.pipe(resizer).pipe(res);
+    } else {
+      res.sendStatus(404);
```
williamchong, line 50: Use default avatar in this case?

williamchong (changes requested):

src/routes/users/getPublicInfo.js
WeiJun0827: BTW, do we have a preferred `minSize` and `maxSIze` for avatar?

src/util/api/users/getPublicInfo.js
```diff
@@ -45,10 +63,35 @@ export function formatUserCivicLikerProperies(id, data) {
...
+    field = 'wallet';
+  } else if (checkCosmosAddressValid(addr, 'like')) {
+    field = 'likeWallet';
+  } else if (checkCosmosAddressValid(addr, 'cosmos')) {
+    field = 'cosmosWallet';
+  } else {
+    throw new ValidationError('Invalid address');
+  }
+  const query = await dbRef.where(field, '==', addr).limit(1).get();
+  if (query.docs.length < 0) return null;
```
nnkken: `<=`?

src/routes/users/getPublicInfo.js
WeiJun0827: Currently, we had handled those who are Liker but don't provide an avatar in `getUserAvatar(id)`:
https://github.com/likecoin/likecoin-api-public/pull/382/files#diff-bee447251ba2e2147edcd9db72fd0bd3026cc5bb4e9c66f8c3431fc7712adcf2R71-R77

And for those who are not Liker, we display avatars generated by dicebear API in liker.land.

Do we prefer to use the default avatar rather than dicebear avatar for web3 users?

nnkken, line 50: Usually we prefer exit fast for this kind of `if` branches, i.e. 
```JS
if (!avatar) {
  res.sendStatus(404);
  // may also need a `return`, depending on how likely we will add things after the whole `if` block in the future
}
// normal logic...
```
This keeps the main branch less indented, also prevents having a long block and then an `else` branch which no body remember what's the context

src/routes/users/getPublicInfo.js
williamchong: maxSize... probably 1920?
minSize > 0 is fine

src/routes/users/getPublicInfo.js
williamchong: Here we only handle likers

WeiJun0827: I see, will return default avatar for this case.

WeiJun0827: On second thought, can someone identify whether the given liker doesn't exist or the liker doesn't set an avatar if this API returns the default avatar? Or it doesn't matter, returning the default avatar for both cases is fine?

williamchong: return 404 on id not exists
return default on id exists but no avatar set

nwingt: Maybe return the default for Liker and dicebear API for wallet without Liker ID ?

williamchong (approved):

PR #391, opened 2022-11-06
✨ Return next new NFT ID for purchase info

nwingt:
For liker.land to retrieve new NFT ID of NFT Class in the new NFT page

### DIFF
--- a/src/util/api/likernft/purchase.js
+++ b/src/util/api/likernft/purchase.js
@@ -95,6 +95,7 @@ export async function getLatestNFTPriceAndInfo(iscnPrefix, classId) {
   let price = -1;
   let isResell = false;
   let nftId;
+  let nextNewNFTId;
   const {
     currentPrice,
     currentBatch,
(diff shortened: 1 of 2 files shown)


williamchong:
What is this ID for? as the comment above sugggest, NFT ID is fetched live when processing purchase transaction to prevent purchase of same price level from blocking each other
https://github.com/likecoin/likecoin-api-public/blob/91ac85cf212afd124819fd474501eb041fc4f7b3/src/util/api/likernft/purchase.js#L344

nwingt:
In the new design, users will go to the NFT details page instead NFT Class details page, so I need to obtain the **FIRST** unsold NFT ID of that NFT Class and redirect users to the proper NFT details page

Maybe there are some situations that haven't been covered

And I guess users may not collect the same NFT ID shown on the NFT details page as someone else has probably collected that
image

williamchong (approved):
Ok if we don't use this ID thinking it is the purchased ID

src/util/api/likernft/purchase.js
```diff
@@ -103,6 +104,7 @@ export async function getLatestNFTPriceAndInfo(iscnPrefix, classId) {
   if (newNftData) {
     price = currentPrice;
     // Do not set NFT ID since another fresh one might be used on purchase instead
```
williamchong: Maybe modify this comment though

williamchong (changes requested):

PR #397, opened 2022-11-18
🏷️ Feature/typescript

williamchong:
Hack performed:
- `any` is used sparsely
- ava hangs on live ts-script compile, so precompile mode is used for testing, code coverage report thus might have offset line
- c8 randomly segfault on ci
- eslint version mismatch with web3.storage, causing chain of eslint dependencies to be upgraded and thus huge diff
- sharp crashed unittest on ci, a `require('sharp')` hack is used in ava.config.cjs
- refactored oembed test when investigating test fail


src/util/api/likernft/index.ts
```diff
@@ -32,10 +32,10 @@ export async function getISCNPrefixByClassId(classId) {
   return decodeURIComponent(doc.id);
 }
 
-
 export async function getNFTClassByClassId(classId) {
   const c = await getNFTQueryClient();
   const client = await c.getQueryClient();
   const res = await client.nft.class(classId);
-  return parseNFTClassDataFields(res);
+  if (!res.class) return new Error('CLASS_NOT_FOUND');
```
nwingt: should throw instead of return?

src/util/oauth/flickr.ts
```diff
@@ -29,8 +28,8 @@ export async function fetchFlickrOAuthInfo(user) {
   };
   const { data } = await axios({
     url: req.url,
-    method: req.method,
-    data: querystring.stringify(oauth.authorize(req)),
+    method: req.method as 'POST',
```
nwingt: Axios should provide the type `Method` or simply use `axios.post()`?

src/util/oauth/flickr.ts
```diff
@@ -55,8 +54,8 @@ export async function fetchFlickrUser(oAuthToken, oAuthTokenSecret, oAuthVerifie
   };
   const { data } = await axios({
     url: req.url,
-    method: req.method,
-    data: querystring.stringify(oauth.authorize(req, token)),
+    method: req.method as 'POST',
+    data: querystring.stringify(oauth.authorize(req, token)as any),
```
nwingt:
```suggestion
    data: querystring.stringify(oauth.authorize(req, token) as any),
```

nwingt (approved):

PR #398, opened 2022-11-23
✨ Add update ts to invalidate avatar cache

williamchong:


WeiJun0827 (commented):
How will `/users/id/:id/avatar` handle the `ts`?

williamchong:
> How will `/users/id/:id/avatar` handle the `ts`?

it won't be handled, it is just for bypassing the existing cache for each ts/canonical

WeiJun0827:
> > How will `/users/id/:id/avatar` handle the `ts`?
> 
> it won't be handled, it is just for bypassing the existing cache for each ts/canonical

Curious about why could we bypass the cache by just adding `tx`?

williamchong:
> > > How will `/users/id/:id/avatar` handle the `ts`?
> > 
> > 
> > it won't be handled, it is just for bypassing the existing cache for each ts/canonical
> 
> Curious about why could we bypass the cache by just adding `tx`?

You can often add random querystring behind any url to bypass cache
Most of the time the cache key is the whole url, so it does not use the same cache when qs is different

WeiJun0827:
I see, thanks!! 🙏🏻

WeiJun0827 (approved):

PR #405, opened 2022-11-28
🔥 Remove social apis

williamchong:
We still use matters on api register, others are removed


nwingt:
We need to hide the UI in legacy like.co first?

PR #420, opened 2023-01-19
✨ Support `memo` for Stripe purchase

WeiJun0827:


src/routes/likernft/fiat/stripe.ts
```diff
@@ -182,6 +185,7 @@ router.post(
         paymentId,
         buyerWallet: wallet,
         classId,
+        memo,
```
williamchong:
```suggestion
        BuyerMemo: memo,
```

src/util/api/likernft/fiat/index.ts
```diff
@@ -128,6 +129,7 @@ export async function processFiatNFTPurchase({
       classId,
       granterWallet: fiatGranterWallet,
       grantedAmount: LIKEPrice,
+      granterMemo: memo,
```
williamchong, line 133: Not sure if we need `grantTxHash` also

williamchong (changes requested):

WeiJun0827:
截圖 2023-01-19 下午8 17 29

PR #422, opened 2023-01-19
✨ Support tx_hash query in nft history

williamchong:
Might be useful if we only want to query price info of a known transaction


WeiJun0827 (commented):
Maybe can add a guard like this?
```
if (nftId && txHash) {
  throw new ValidationError('xxx');
}
```

WeiJun0827 (approved):

PR #429, opened 2023-02-13
👽️ Include `createdTotalSales` in user stats

WeiJun0827:


src/util/api/likernft/user.ts
```diff
@@ -43,7 +44,9 @@ export async function getUserStat(wallet) {
     collected_classes: collectedClasses,
     created_count: createdClassCount,
     collector_count: createdCollectorCount,
+    total_sales: totalSalesInNanolike,
   } = userStat;
+  const createdTotalSales = new BigNumber(totalSalesInNanolike).shiftedBy(-9).toNumber();
```
williamchong: maybe round to 2dp or integer?

williamchong (changes requested):

src/util/api/likernft/user.ts
WeiJun0827: Yeah, can do. I was thinking leaving the decision to front-end to make?

williamchong: you can't expect to send a bignumber with `number`, if you want frontend to make decision it needs to be `string`

WeiJun0827: I see, I'll just return an integer.

WeiJun0827: I see, I'll just return an integer.

PR #432, opened 2023-02-21
✨ Marketplace listing supports paying by Stripe

WeiJun0827:


src/routes/likernft/fiat/stripe.ts
```diff
@@ -60,18 +61,34 @@ router.get(
...
       const totalPrice = price + gasFee;
       const fiatPriceString = await getFiatPriceStringForLIKE(totalPrice);
-      res.json({
+      const payload = {
         LIKEPrice: totalPrice,
         fiatPrice: Number(fiatPriceString),
         fiatPriceString,
-      });
+        isListing,
+        listingInfo: {},
```
williamchong: Prefer to be null if not listing

src/routes/likernft/fiat/stripe.ts
```diff
@@ -88,22 +105,33 @@ router.post(
...
+        promises.push(getLatestNFTPriceAndInfo(iscnPrefix, classId));
+      }
+      const [{ metadata }, info] = await Promise.all(promises) as any;
       let { name = '', description = '' } = metadata;
       const { image } = metadata;
       const gasFee = getGasPrice();
+      let price = 0;
+      if (isListing) {
+        const listingInfo = info;
+        const targetListing = listingInfo.find((l) => l.nftId === nftId && l.seller === seller);
```
williamchong: Just use the `classId/nftId` query instead of using `.find`
https://github.com/likecoin/likecoin-chain/blob/master/proto/likechain/likenft/v1/query.proto#L109

src/util/api/likernft/fiat/index.ts
```diff
@@ -123,15 +132,27 @@ export async function processFiatNFTPurchase({
...
-    res = await processNFTPurchase({
-      buyerWallet: likeWallet,
-      iscnPrefix,
-      classId,
-      granterWallet: fiatGranterWallet,
-      grantedAmount: LIKEPrice,
-      grantTxHash: paymentId,
-      granterMemo: memo,
-    }, req);
+    if (listingNftId && seller) {
```
williamchong: I think we want to set a `isListing` flag as input param that explicity state we want to use listing
The newly added params `nftId` and `seller` both can have meaning outside listing, we may have other use for them later

src/util/api/likernft/listing.ts
```diff
@@ -0,0 +1,136 @@
...
+  buyerWallet: string;
+  classId: string;
+  nftId: string;
+  sellerWallet: string;
+  priceInLIKE: number;
+  memo?: string;
+}) {
+  const { client, wallet } = await getLikerNFTFiatSigningClientAndWallet();
+  const fiatWallet = wallet.address;
+  const amount = new BigNumber(priceInLIKE).shiftedBy(9).toFixed(0);
```
williamchong: Can we pass the exact amount (string) in nanolike we obtain in API instead? Any non-integer-LIKE might break this API due to precision error

williamchong (changes requested):

src/util/api/likernft/listing.ts
```diff
@@ -0,0 +1,136 @@
...
+  nftId: string;
+  sellerWallet: string;
+  priceInLIKE: number;
+  memo?: string;
+}, req) {
+  const listingInfo = await fetchNFTListingInfo(classId);
+  const listing = listingInfo.find((l) => l.nftId === nftId && l.seller === sellerWallet);
+  if (!listing) throw new ValidationError('LISTING_NOT_FOUND');
+  const { price: actualNftPrice } = listing;
+  if (priceInLIKE < actualNftPrice) throw new ValidationError('LISTING_PRICE_NOT_MATCH');
```
williamchong: Should use the original nanolike string in listing for comparison to prevent precision error

williamchong (changes requested):

src/util/api/likernft/fiat/index.ts
WeiJun0827: I'll store the `isListing` flag in the DB, so the caller of this function can explicitly get the flag from DB and pass it to this function.

src/util/api/likernft/listing.ts
```diff
@@ -42,43 +42,41 @@ function formateListingInfo(info: {
 export async function fetchNFTListingInfo(classId: string) {
   const { data } = await axios.get(`${COSMOS_LCD_INDEXER_ENDPOINT}/likechain/likenft/v1/listings/${classId}`);
   const info = data.listings
-    .map(formateListingInfo)
+    .map(formatListingInfo)
```
williamchong: Strange to have different return format of `fetchNFTListingInfo` and `fetchNFTListingInfoByNFTId` differ by `formatListingInfo`. Can we not call `formatListingInfo` here and call it where we use `fetchNFTListingInfo`?

src/util/api/likernft/listing.ts
```diff
@@ -109,19 +107,19 @@ export async function processNFTBuyListing({
...
-  const listingInfo = await fetchNFTListingInfo(classId);
-  const listing = listingInfo.find((l) => l.nftId === nftId && l.seller === sellerWallet);
-  if (!listing) throw new ValidationError('LISTING_NOT_FOUND');
-  const { price: actualNftPrice } = listing;
-  if (priceInLIKE < actualNftPrice) throw new ValidationError('LISTING_PRICE_NOT_MATCH');
+  const listingInfo = await fetchNFTListingInfoByNFTId(classId, nftId);
+  if (!listingInfo) throw new ValidationError('LISTING_NOT_FOUND');
+  const { price: actualNftPrice } = listingInfo;
+  const actualPriceInLIKE = new BigNumber(actualNftPrice).shiftedBy(-9).toNumber();
+  if (priceInLIKE < actualPriceInLIKE) throw new ValidationError('LISTING_PRICE_NOT_MATCH');
```
williamchong: Better do this comparision with `bignumber`

williamchong (changes requested):

williamchong (approved):

src/routes/likernft/fiat/stripe.ts
```diff
@@ -88,22 +116,32 @@ router.post(
...
       const gasFee = getGasPrice();
+      let price = 0;
+      if (isListing) {
+        const listingInfo = formatListingInfo(info);
+        if (!listingInfo || listingInfo.seller !== seller) throw new ValidationError('LISTING_NOT_FOUND');
+        ({ price } = listingInfo);
+      } else {
+        const purchaseInfo = info;
+        ({ price } = purchaseInfo);
+      }
```
nnkken: I suggest putting these logics into `then` block, so there would be less distracting `if` branches
```JS
if (isListing) {
  promises.push(
    fetchNFTListingInfoByNFTId(classId, nftId)
      .then((info) => {
        const listingInfo = formatListingInfo(info);
        if (!listingInfo || listingInfo.seller !== seller) throw new ValidationError('LISTING_NOT_FOUND');
        const { price } = listingInfo;
        return price;
      }),
  ); 
} else {
  promises.push(
    getLatestNFTPriceAndInfo(iscnPrefix, classId)
      .then(({ price }) => price),
  );
}
const [{ metadata }, price] = await Promise.all(promises) as any;
```

src/util/api/likernft/fiat/index.ts
```diff
@@ -42,7 +43,7 @@ async function getLIKEPrice() {
 export async function getFiatPriceStringForLIKE(LIKE, { buffer = 0.1 } = {}) {
   const rate = await getLIKEPrice();
   const price = new BigNumber(LIKE).multipliedBy(rate).multipliedBy(1 + buffer);
-  const total = price.plus(LIKER_NFT_FIAT_FEE_USD).toFixed(2);
+  const total = price.plus(LIKER_NFT_FIAT_FEE_USD).toFixed(2, 2); // ROUND_CEIL
```
nnkken: Simply use `BigNumber.ROUND_CEIL` directly

src/routes/likernft/fiat/stripe.ts
WeiJun0827: Much clearer, great insight, thanks!

nnkken (approved):

PR #439, opened 2023-03-09
✨ Add nft subscription routes

williamchong:


src/routes/likernft/subscription/mint.ts
```diff
@@ -418,14 +426,20 @@ router.get(
...
+      const { signature, publicKey, message } = req.body;
+      if (!checkCosmosSignPayload({
+        signature, publicKey, message, inputWallet: wallet as string, action: 'list_mint',
+      })) {
+        throw new ValidationError('INVALID_SIGN', 401);
+      }
       if (!isValidLikeAddress(wallet)) throw new ValidationError('INVALID_WALLET');
       const list = await getAllMintTransaction(wallet as string);
-      res.json({ list });
+      res.json({ list: list.map((i) => i) });
```
nnkken: Why?
If you want a copy (which I don't know why needed in `res.json`), simply do `[...list]` is more straight forward

williamchong: it is for the `filterNFTSubscriptionMintStatus` function in the later commit, which I haven't wrote in this commit yet

src/util/api/likernft/subscription/stripe.ts
```diff
@@ -154,6 +154,7 @@ export async function processStripeNFTSubscriptionInvoice(
       currentBillingPeriod: 'month',
       currentPeriodStart,
       currentPeriodEnd,
+      currentPeriodMints: 0,
```
nnkken, line 157: should also init for `totalMints`?

src/util/api/likernft/subscription/index.ts
```diff
@@ -0,0 +1,141 @@
...
+  'new',
+  'arweave',
+  'iscn',
+  'nftCover',
+  'nftClass',
+  'nftMint',
+  'done',
+];
+
+export function createMintAutheticationToken(sessionId: string): string {
```
WeiJun0827:
```suggestion
export function createMintAuthenticationToken(sessionId: string): string {
```

src/routes/likernft/subscription/status.ts
```diff
@@ -0,0 +1,24 @@
+import { Router } from 'express';
+
+import { likeNFTSubscriptionUserCollection } from '../../../util/firebase';
+
+const router = Router();
+
+router.get('/', async (req, res, next) => {
+  try {
+    const { wallet } = req.query;
```
WeiJun0827, line 11: Should we guard with `isValidLikeAddress(wallet)` here?

src/routes/likernft/subscription/status.ts
```diff
@@ -0,0 +1,24 @@
...
+    const doc = await likeNFTSubscriptionUserCollection.doc(wallet).get();
+    if (!doc.data()) {
+      res.status(404).send('PAYMENT_ID_NOT_FOUND');
+      return;
+    }
+    const { currentPeriodEnd, currentPeriodStart } = doc.data();
+    const tsNow = Date.now() / 1000;
+    const isActive = currentPeriodStart < tsNow && currentPeriodEnd > tsNow;
+    res.json({ isActive });
```
WeiJun0827: Is it suitable to reuse `checkUserIsActiveNFTSubscriber()` here?

src/util/api/likernft/subscription/stripe.ts
williamchong: In cases where user cancel then subscribe again, init-ing `totalMints` here would overwrite the old data
So I rely on a behaviour where calling `Field.increment(1)` on undefined field would set it to `1`
But on second init-ing `totalMints` here might be a better idea for cleanness and just define it as per-subscription stat
We can always check log for lifetime value

WeiJun0827 (approved):

PR #440, opened 2023-03-11
✨ Support buying listing P.NFT by Stripe

WeiJun0827:


src/util/api/likernft/purchase.ts
```diff
@@ -136,6 +136,15 @@ export async function getLatestNFTPriceAndInfo(iscnPrefix, classId) {
   };
 }
 
+export async function softGetLatestNFTPriceAndInfo(iscnPrefix, classId) {
```
williamchong, line 139: @nwingt any good naming convention for try catch wrapped functions?

src/middleware/likernft.ts
```diff
@@ -27,4 +30,16 @@ export const fetchISCNPrefixAndClassId = async (req, res, next) => {
   }
 };
 
+export const fetchISCNPrefixFromChain = async (req, res, next) => {
+  try {
+    const { class_id: classId } = req.query;
+    if (!classId) throw new ValidationError('MISSING_CLASS_ID');
+    const iscnPrefix = await getISCNPrefixByClassIdFromChain(classId);
```
williamchong, line 37: Should handle case where is minted via account, not iscn. in this case `getISCNPrefixByClassIdFromChain` return undefined
Should throw error in this case?

src/util/api/likernft/index.ts
```diff
@@ -32,6 +32,14 @@ export async function getISCNPrefixByClassId(classId) {
   return decodeURIComponent(doc.id);
 }
 
+export async function getISCNPrefixByClassIdFromChain(classId) {
+  const data = await getNFTClassDataById(classId);
+  if (!data) {
+    throw new ValidationError('NFT_CLASS_NOT_FOUND', 404);
+  }
+  return data.data.parent.iscnIdPrefix;
```
williamchong: Note that if the nft is minted via account parent, this can be undefined

src/routes/likernft/fiat/stripe.ts
```diff
@@ -108,33 +116,38 @@ router.get(
...
+      promises.push(getPriceInfoPromise);
+      const [metadata, priceInfo] = await Promise.all(promises) as any;
+      if (!priceInfo) throw new ValidationError('NFT_PRICE_NOT_FOUND');
+      const { price } = priceInfo;
+      let {
+        name = '',
+        description = '',
+        image,
+      } = metadata;
+      image = image ? parseImageURLFromMetadata(image) : `https://${API_EXTERNAL_HOSTNAME}/likernft/metadata/image/class_${classId}?size=${DEFAULT_NFT_IMAGE_SIZE}`;
```
williamchong: I think we can use metadata to judge which image should we use
- is writing nft but not custom image: `https://${API_EXTERNAL_HOSTNAME}/likernft/metadata/image/class_${classId}?size=${DEFAULT_NFT_IMAGE_SIZE}`
- is writing nft but custom image, or not writing nft:
`parseImageURLFromMetadata(image)`
- we also want a static nft image(the grey one used in likerland?) as fallback if nothing exists

williamchong (changes requested):

PR #444, opened 2023-03-30
🚸 Self-recover missing NFT sold info in DB

WeiJun0827:


src/util/api/likernft/purchase.ts
```diff
@@ -346,6 +352,84 @@ export async function handleNFTPurchaseTransaction({
...
+    grantTxHash,
+    granterMemo,
+    memo,
+    price,
+    classId,
+    nftId,
+    timestamp,
+    fromWallet: seller,
+    toWallet: owner,
+    granterWallet,
```
WeiJun0827, line 431: Note that we don't provide `sellerWallet, sellerLIKE, stakeholderWallets, stakeholderLIKEs` which appear in the successful purchase flow. Not sure if it's necessary for now/future.

src/util/api/likernft/purchase.ts
```diff
@@ -553,7 +638,24 @@ export async function processNFTPurchase({
...
+          classRef,
+          classId,
+          nftId,
+          granterWallet,
+          grantTxHash,
+          granterMemo,
+        });
+      }
       t.update(nftRef, { isProcessing: false });
+      return shouldUpdateSoldNFT && retryTimes < 1;
```
WeiJun0827, line 660: If the flow is acceptable, probably can move the `MAX_NFT_PURCHASE_RETRY_TIMES` to config.js

src/util/api/likernft/purchase.ts
```diff
@@ -346,6 +352,84 @@ export async function handleNFTPurchaseTransaction({
...
+    'pagination.limit': 1,
+    'pagination.reverse': true,
+  };
+  const { data: { events } } = await axios.get(`${COSMOS_LCD_INDEXER_ENDPOINT}/likechain/likenft/v1/event`, { params });
+  return events.length ? events[0] : null;
+}
+
+function formatNFTEvent(event: any) {
+  const timestamp = new Date(event.timestamp).getTime();
+  const price = Number(new BigNumber(event.price).shiftedBy(-9).toFixed(0));
```
williamchong, line 372: Why can we do `toFixed(0)` here? Price can be non-integer

williamchong (changes requested):

src/util/api/likernft/purchase.ts
WeiJun0827: The `event.price` includes tx fee (e.g. `8002000000 nanolike`), but the price in FIrestore is an integer that excludes tx fee. Given the condition of NFT sold by API wallet, the tx price queried from `/event` API should be formatted to an integer after excluding the tx fee?

williamchong: Please add a comment to explain this situation

nnkken (approved):

williamchong (approved):

PR #445, opened 2023-04-13
[TAS-821] ✨ Add nft collection API

williamchong:
Implement a user curated list of nft collection
for now we would only have book collection which is listed on book store
in the future we might support my favorite,to-read etc function


src/routes/likernft/collection/index.ts
```diff
@@ -0,0 +1,110 @@
...
+
+router.get('/collection/:collectionId', jwtOptionalAuth('read:nftcollection'), async (req, res, next) => {
+  try {
+    const wallet = req.user?.wallet;
+    const { collectionId } = req.params;
+    const doc = await likeNFTCollectionCollection.doc(collectionId).get();
+    const docData = doc.data();
+    if (!docData) {
+      throw new ValidationError('COLLECTION_NOT_FOUND', 404);
+    }
```
WeiJun0827: Should use `getNFTCollectionById` here?

PR #447, opened 2023-04-18
✨ Auto-send email for pending claim Stripe purchase

WeiJun0827:


src/util/api/likernft/fiat/stripe.ts
```diff
@@ -193,5 +195,17 @@ export async function processStripeFiatNFTPurchase(session, req) {
       console.error(err);
     }
   }
+  if (metadata.isPendingClaim) {
+    try {
+      const iscnData = await getNFTISCNData(iscnPrefix);
+      const className = iscnData.data?.contentMetadata.name;
+      await sendPendingClaimEmail(customer.email, classId, className);
```
williamchong: We should send email first, and state the result of email sending or not in the slack message above
this is to allow easier customer service follow up (i.e. don't send again if already sent automatically, handle with extra effort if sending email failed)

williamchong (changes requested):

src/util/api/likernft/fiat/stripe.ts
```diff
@@ -150,6 +165,9 @@ export async function processStripeFiatNFTPurchase(session, req) {
       }
       words.push(isPendingClaim ? 'An unclaimed' : 'A');
       words.push('NFT is bought');
+      if (isEmailSent !== null) {
```
williamchong: Use `const { isPendingClaim = false } = metadata; if (isPendingClaim)` for the condition check, then we can just init `isPendingClaimEmailSent: boolean = false` above?
Please use the name `isPendingClaimEmailSent` to clear its purpose

williamchong (changes requested):

williamchong (approved):

src/util/ses.ts
```diff
@@ -116,3 +116,68 @@ export async function sendInvitationEmail(res, { email, referrerId, referrer })
...
+          <p><strong>Method 2: Install Keplr</strong></p>
+          <p>If you want to receive the Writing NFT with a new LikeCoin address, you can refer to the following steps and install the Keplr wallet 
+            on the Chrome browser.</p>
+          <p><a href="https://docs.like.co/general-guides/wallet/keplr">How to install Keplr and use LikeCoin</a></p>
+          <p>After completing the installation, you will be able to view your Writing NFT collection on the <a href="https://liker.land/">Liker Land website</a>.</p>
+          <p>＋＋＋＋</p>
+          <p>Please reply to this email with your LikeCoin wallet address.</p>
+          <p>If you encounter any problems, please contact <a href="https://go.crisp.chat/chat/embed/?website_id=5c009125-5863-4059-ba65-43f177ca33f7">Liker
+            Land customer service</a>.</p>
+          <p>Thank you for supporting our creation.</p>
```
williamchong:
```suggestion
          <p>Thank you for supporting creativity.</p>
```

williamchong (changes requested):

williamchong (approved):

PR #454, opened 2023-04-29
✨ Support auto-claiming flow for Stripe purchase w/o providing wallet

WeiJun0827:
Drafted the claiming steps for the Chinese part of the letter, and will update the English part if it is finalized.


src/routes/likernft/fiat/stripe.ts
```diff
@@ -302,4 +295,76 @@ router.get(
   },
 );
 
+router.get(
+  '/pending-count',
```
williamchong:
```suggestion
  '/pending/count',
```

src/routes/likernft/fiat/stripe.ts
```diff
@@ -302,4 +295,76 @@ router.get(
...
+        .get();
+      res.json({ count: snapshot.docs.length });
+    } catch (err) {
+      next(err);
+    }
+  },
+);
+
+router.post(
+  '/claim',
```
williamchong:
```suggestion
  '/pending/claim',
```

src/util/api/likernft/fiat/stripe.ts
```diff
@@ -78,11 +88,16 @@ export async function processStripeFiatNFTPurchase(session, req) {
...
+  if (!isWalletProvided && email) {
+    wallet = await findWalletWithVerifiedEmail(email);
+  }
+  const isPendingClaim = !wallet;
+  let claimToken;
   if (isPendingClaim) {
-    verifiedWallet = await findWalletWithVerifiedEmail(customer.email);
-    wallet = verifiedWallet || wallet;
+    wallet = LIKER_NFT_PENDING_CLAIM_ADDRESS;
+    claimToken = uuidv4();
```
williamchong: Prefer using random at least 32 bytes converted to hex or base64

src/util/api/likernft/fiat/stripe.ts
```diff
@@ -155,25 +172,30 @@ export async function processStripeFiatNFTPurchase(session, req) {
     sessionId,
   });
   let isEmailSent = false;
-  if (isPendingClaim) {
-    const { email } = customer;
+  if (!TEST_MODE && !isWalletProvided) {
```
williamchong: Why not send in `TEST_MODE` ?

src/util/api/likernft/fiat/index.ts
```diff
@@ -200,7 +203,9 @@ export async function processFiatNFTPurchase({
     transactionHash,
     nftId,
     actualNftPrice,
-    status: 'done',
+    claimToken: claimToken || undefined,
+    status: claimToken ? 'pending' : 'done',
```
williamchong:
```suggestion
    status: claimToken ? 'pendingClaim' : 'done',
```

src/routes/likernft/fiat/stripe.ts
```diff
@@ -302,4 +295,76 @@ router.get(
...
 
+router.get(
+  '/pending-count',
+  async (req, res, next) => {
+    try {
+      const { email } = req.query;
+      if (!email) throw new ValidationError('EMAIL_NEEDED');
+      const snapshot = await likeNFTFiatCollection
+        .where('email', '==', email)
+        .where('status', '==', 'pending')
```
williamchong:
```suggestion
        .where('status', '==', 'pendingClaim')
```

src/routes/likernft/fiat/stripe.ts
```diff
@@ -302,4 +295,76 @@ router.get(
...
+        const doc = await t.get(ref);
+        if (!doc.exists) throw new ValidationError('PAYMENT_ID_NOT_FOUND', 404);
+        const {
+          classId,
+          nftId,
+          status,
+          claimToken,
+        } = doc.data();
+        if (claimToken !== token) throw new ValidationError('INVALID_TOKEN', 403);
+        if (status !== 'pending') throw new ValidationError('ALREADY_CLAIMED_OR_CLAIM_ERROR', 409);
```
williamchong:
```suggestion
        if (status !== 'pendingClaim') throw new ValidationError('ALREADY_CLAIMED_OR_CLAIM_ERROR', 409);
```

src/routes/likernft/fiat/stripe.ts
```diff
@@ -302,4 +295,76 @@ router.get(
...
+
+        const {
+          client,
+          wallet: senderAccount,
+        } = await getLikerNFTPendingClaimSigningClientAndWallet();
+        const txRes = await client.sendNFTs(
```
williamchong: Should update `status` for the tx to prevent race condition, otherwise during `getLikerNFTPendingClaimSigningClientAndWallet` before `t.update(ref`, it is possible multiple tx can be triggered since `status` is still `pending`
In this case any error should reset status, or update status to error

williamchong (changes requested):

src/util/api/likernft/fiat/stripe.ts
WeiJun0827: Because we don't load config/aws.json in `TEST_MODE`?
https://github.com/likecoin/likecoin-api-public/blob/master/src/util/ses.ts#L6

williamchong: Maybe the guard should be inside the functions in `ses.ts` instead of all calling functions

src/routes/likernft/fiat/stripe.ts
```diff
@@ -302,4 +295,107 @@ router.get(
...
+      if (!token) throw new ValidationError('TOKEN_NEEDED');
+
+      const ref = likeNFTFiatCollection.doc(paymentId);
+      let classId;
+      let nftId;
+      await db.runTransaction(async (t) => {
+        const doc = await t.get(ref);
+        if (!doc.exists) throw new ValidationError('PAYMENT_ID_NOT_FOUND', 404);
+        const { status, claimToken } = doc.data();
+        ({ classId, nftId } = doc.data());
```
williamchong: We can return `{ classId, nftId }` in this transaction instead of using `let`

src/routes/likernft/fiat/stripe.ts
```diff
@@ -302,4 +295,107 @@ router.get(
...
+          // eslint-disable-next-line no-console
+          console.log(`NFT Claim already handled ${paymentId}`);
+          publisher.publish(PUBSUB_TOPIC_MISC, req, {
+            logType: 'LikerNFTFiatClaimAlreadyHandled',
+            paymentId,
+            classId,
+            nftId,
+            receiverWallet,
+          });
+          throw new ValidationError('NFT_CLAIM_ALREADY_HANDLED', 409);
```
williamchong: Should fire pubsub (generally anything with side effect) outside of transaction, since transaction might be re-run if the `get` data is changed

src/routes/likernft/fiat/stripe.ts
```diff
@@ -302,4 +295,107 @@ router.get(
...
+        const {
+          client,
+          wallet: senderAccount,
+        } = await getLikerNFTPendingClaimSigningClientAndWallet();
+        txRes = await client.sendNFTs(
+          senderAccount.address,
+          receiverWallet as string,
+          classId,
+          [nftId],
+        );
```
williamchong: Should we use `sendTransactionWithSequence` here? possible sequence race condition

src/routes/likernft/fiat/stripe.ts
```diff
@@ -302,4 +295,107 @@ router.get(
...
+        await ref.update({
+          status: 'error',
+          error,
+          errorMessage,
+          errorStack,
+        });
+        throw err;
+      }
+      const { transactionHash: txHash, code } = txRes as DeliverTxResponse;
+      if (code) throw new ValidationError(`TX_${txHash}_FAILED_WITH_CODE_${code}`);
```
williamchong: Is this a `LikerNFTFiatClaimError` too? Should handle like above

src/util/api/likernft/fiat/index.ts
```diff
@@ -1,8 +1,9 @@
 import axios from 'axios';
 import BigNumber from 'bignumber.js';
-
 import { DeliverTxResponse } from '@cosmjs/stargate';
 import LRU from 'lru-cache';
+import uuidv4 from 'uuid/v4';
```
williamchong: Can remove if we are not using this now

williamchong (changes requested):

src/routes/likernft/fiat/stripe.ts
WeiJun0827: Will skip this pubsub and let the pubsub with `logType: 'LikerNFTFiatClaimError'` in catch scope handle.

WeiJun0827: Oh I see, that's the mechanism of Firestore transaction. Thanks!

WeiJun0827: By the way, is it necessary to add a log here? I copied the code from `processFiatNFTPurchase()`, but not sure if it's OK to just return a `409` without additional log-out.

williamchong: What do you mean by log-out?

WeiJun0827: Like `console.log` and pubsub here?

williamchong: Good to have consistent logging for all fiat NFT purchase error, better keep the logs
but seems like a `console.error` than `console.log` to me

williamchong (approved):
Missing english email before merge

PR #460, opened 2023-05-17
✨ Support purchasing multiple W.NFT in a batch

WeiJun0827:
I think the simplest way of the E2E test is hosting rinkeby.liker.land and API server locally and hardcoding [postNFTPurchase()](https://github.com/likecoin/liker-land/blob/master/src/util/api/index.js#L233)'s `class_id` with an array.


src/util/NFTValidationError.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class NFTValidationError extends ValidationError {
+  status = 400;
+
+  nftId: string;
+
+  constructor({
+    message,
+    status = 400,
+    nftId = '',
```
williamchong: Should we support custom data payload in `ValidationError`, instead of making `NFTValidationError`?

src/routes/likernft/purchase.ts
```diff
@@ -72,61 +78,58 @@ router.post(
...
       }, req);
       res.json({
-        txHash: transactionHash,
-        iscnId: iscnPrefix,
-        classId,
-        nftId,
-        nftPrice: actualNftPrice,
-        gasFee: actualGasFee,
+        txHash,
+        purchaseInfoList: purchaseInfoList.map((info) => ({
```
williamchong:
```suggestion
        purchased: purchaseInfoList.map((info) => ({
```

src/util/api/likernft/purchase.ts
```diff
@@ -554,121 +558,126 @@ async function updateDocsForMissingSoldNFT(t, {
...
+          t,
+        })
+      )));
 
       // eslint-disable-next-line no-underscore-dangle
       const shouldUpdateSoldNFT = (
-        err instanceof ValidationError
+        err instanceof NFTValidationError
         && err.message === 'NFT_NOT_OWNED_BY_API_WALLET'
+        && err.nftId
```
williamchong: Is it possible that multiple nfts disappeared, would a failed transaction contains info of all these nfts (instead of just the first one), and can we try all of them together here?

src/util/api/likernft/purchase.ts
```diff
@@ -554,121 +558,126 @@ async function updateDocsForMissingSoldNFT(t, {
...
     const shouldRetryPurchase = await db.runTransaction(async (t) => {
-      await resetLockedPurchaseDocs({
-        iscnPrefix,
-        classId,
-        nftId,
-        currentBatch,
-        t,
-      });
+      await Promise.all(classIds.map((_, i) => (
+        resetLockedPurchaseDocs({
```
williamchong, line 674: Does firestore transaction has a limitation of number of documents locked? Please guard nft length in API for the limit

src/util/api/likernft/purchase.ts
```diff
@@ -554,121 +558,126 @@ async function updateDocsForMissingSoldNFT(t, {
...
+      });
     });
 
     await db.runTransaction(async (t) => {
-      await updateDocsForSuccessPurchase(t, {
-        iscnPrefix,
-        classId,
-        nftId,
-        sellerWallet,
+      await Promise.all(purchaseInfoList.map((info) => updateDocsForSuccessPurchase(t, {
```
williamchong: I don't think we should use `Promise.all` inside transaction since ordering here can be undefined, prefer awaiting 1 by 1

src/util/api/likernft/purchase.ts
```diff
@@ -554,121 +558,126 @@ async function updateDocsForMissingSoldNFT(t, {
...
-  const {
-    nextNewNFTId: nftId,
-    price: nftPrice,
-    currentBatch,
-  } = priceInfo;
-  if (nftPrice > grantedAmount) {
+  const priceInfoList = await db.runTransaction((t) => (
+    Promise.all(classIds.map((_, i) => (
+      getPriceInfoAndLockDocs(iscnPrefixes[i], classIds[i], t)
+    )))
```
williamchong: I don't think we should use `Promise.all` inside transaction since ordering here can be undefined, prefer awaiting 1 by 1

williamchong (changes requested):

src/util/api/likernft/purchase.ts
WeiJun0827: Should test if the blockchain could detect multiple ownership errors from the msg array and return feedback in `rawLog`.

src/util/api/likernft/purchase.ts
WeiJun0827: Wonder what side effects would happen if we update docs concurrently with `Promise.all` here?

src/util/api/likernft/purchase.ts
WeiJun0827: The doc says `Each transaction or batch of writes can write to a maximum of 500 documents.`([source](https://firebase.google.com/docs/firestore/manage-data/transactions)). Consider that `updateDocsForSuccessPurchase` will update 3 docs(ISCN, class, NFT) and create one doc(txHash), 500 / 4 = 125. How about setting the limitation of the maximum class number to 100 in a single request?

src/util/api/likernft/purchase.ts
williamchong: Actually i don't know, but firestore transaction requires all get() happen before update(), so if we dont have this guaranteed now then you might have to rewrite this part

WeiJun0827: I see, may keep it sequentially for now.

src/util/api/likernft/purchase.ts
williamchong: Yes that would work

src/util/api/likernft/purchase.ts
williamchong: Note that it is all get() before any update() so if you call get then update in a function for each nft class, we might have to refactor and split the function

WeiJun0827: I see, so awaiting 1 by 1 means reading all necessary docs in a `Promise.all`, then updating docs later?
In that case, we would have the process like this:
- a `Promise.all` for batch reading starts
  - first async `t.get` finished
  - ...
  - last async `t.get` finished
- the `Promise.all` completed
- sync processing data for multiple docs
- sync assign `t.update` for all docs
- async  `t.commit` under the `db.runTransaction` triggered 


But what's the difference between the above 1 by 1 case and the original code? Here's the pseudo-process of original code in my mind:
- a `Promise.all` for batch reading and labeling what to write starts
  - all async `getAndUpdate` launched
    - async `t.get` finished
    - sync processing data
    - sync assign `t.update`
  - first `readAndWriteFunc` completed
  - ...
  - last `readAndWriteFunc` completed 
- the `Promise.all` completed
- async `t.commit` under the `db.runTransaction` triggered 

Since the actual update action happens at the last line of both the above process at once. Would we keep the `t.get` and `t.update` in the same function to make the code more structured?

Here's an example of using `Promise.all` to handle a `getAndUpdate`:
https://stackoverflow.com/a/48310739/7978205

williamchong: `Promise.all` sounds like a bad idea overall in transaction, but if it is only for `get()`s and if we guarantee the order of write operation, I think we can use it for better performance (as long as it works in firestore)

> a Promise.all for batch reading starts
> first async t.get finished
> ...
> last async t.get finished
> the Promise.all completed
> sync processing data for multiple docs
> sync assign t.update for all docs
> async t.commit under the db.runTransaction triggered

so I think what you describe is what we want.
Do any function we are calling in `Promise.all()` involves a `get()` and `update()` inside a same function?

williamchong: > But what's the difference between the above 1 by 1 case and the original code? Here's the pseudo-process of original code in my mind:

Let say there is `[A, B]` in `Promise.all`, and A and B both contains `get()` and `update()`
Do we have any guarantee (code-wise) that `B.get()` will always finish before `A.update()` is called?
If not it is problematic

WeiJun0827: Since the docs to be updated are independent of each other for different classes, is the order of write operation critical here? I though they all success/fail at once when `t.commit()` is triggered.

williamchong: https://firebase.google.com/docs/firestore/manage-data/transactions?hl=zh-tw#transactions
Read must come before write in a same transaction

WeiJun0827: I think the write action is done at the line `this.commit()` inside the `db.runTransaction()` but not the line `t.update()`? The `t.update` is just labeling the doc needs to be updated later. If the call stack has reached the `db.runTransaction()` then all `t.get()` should already be finished.
I tried this snippet and it seems OK:
```
async function readAndWrite(id, t) {
  const ref = collection.doc(id);
  const doc = await t.get(ref);
  const { name } = doc.data();
  t.update(ref, { greeting: `Hello, ${name}` });
}

async function run() {
  const ids = ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J'];
  await db.runTransaction((t) => Promise.all(ids.map((id) => t.create(collection.doc(id), { name: id }))));
  console.log('Documents created.');
  await db.runTransaction((t) => Promise.all(ids.map((id) => readAndWrite(id, t))));
  console.log('Read and write done.');
}

run();
```

WeiJun0827: Seems fail if I add this part at the beginning of `readAndWrite`
```
const randomSleep = Math.random() >= 0.5;
if (randomSleep) {
    await new Promise(resolve => setTimeout(resolve, 5000));
}
```
Firestore doesn't work as I though before. Should split the read and write later.

src/util/api/likernft/purchase.ts
WeiJun0827: Seems we can only get one error a time.
Tested [this case](https://node.testnet.like.co/cosmos/tx/v1beta1/txs/B3616FB1C4045957E7FCE952A9506A14187B6B5BBCDDD2D21E7B9C338A2CBE9D).
Where both NFTs are sold in [previous purchase](https://node.testnet.like.co/cosmos/tx/v1beta1/txs/0CBA7D8371EF280CA067D598829D59E1CC618F6B8B7B5C54D053CA45659EF95C), but we can only see one unauthorized error in `raw_log`. 
I guess the chain throws an error as soon as the ownership mismatch is detected and doesn't check the other possible error.

williamchong:
One quick question before review: Can indexer parse this kind of multiple purchase events in one tx?

WeiJun0827:
> One quick question before review: Can indexer parse this kind of multiple purchase events in one tx?

https://node.testnet.like.co/cosmos/tx/v1beta1/txs/0CBA7D8371EF280CA067D598829D59E1CC618F6B8B7B5C54D053CA45659EF95C
Seems it works OK with `nft_event`, but we have some problems with `nft_income` since the parsing logic doesn't expect multiple purchases for now. Should further rewrite paring logic.

截圖 2023-06-04 下午1 40 58
截圖 2023-06-04 下午4 40 59

src/middleware/likernft.ts
```diff
@@ -30,6 +30,21 @@ export const fetchISCNPrefixAndClassId = async (req, res, next) => {
   }
 };
 
+export const fetchISCNPrefixes = async (req, res, next) => {
+  try {
+    const { class_id: classId } = req.query;
+    if (!classId) throw new ValidationError('MISSING_ISCN_OR_CLASS_ID');
+    const classIds = Array.isArray(classId) ? classId : [classId];
+    if (classIds.length > 100) throw new ValidationError('CLASS_NUMBER_EXCESS_100', 422);
```
williamchong: Please name the number 100 with a const

src/util/api/likernft/purchase.ts
```diff
@@ -554,121 +561,138 @@ async function updateDocsForMissingSoldNFT(t, {
...
+  const iscnDataList = await Promise.all(iscnPrefixes.map(async (iscnPrefix) => {
+    const iscnData = await getNFTISCNData(iscnPrefix); // always fetch from prefix
+    if (!iscnData) throw new ValidationError('ISCN_DATA_NOT_FOUND');
+    return iscnData;
+  }));
 
-  const memo = await generateCustomMemo(iscnPrefix, classId, buyerWallet);
+  const memo = classIds.length === 1
+    ? await generateCustomMemo(iscnPrefixes[0], classIds[0], buyerWallet)
+    : '(multiple purchase)';
```
williamchong:
```suggestion
    : '(multiple purchases)';
```

williamchong (commented):
Please also update indexer's extractor, and likerland's memo display logic to support multiple purchase

src/middleware/likernft.ts
WeiJun0827: Moved to config.js, is it OK?

williamchong (approved):

PR #461, opened 2023-05-18
✨ Support multiple prices for book store item

williamchong:
price => prices for all books


src/routes/likernft/book/purchase.ts
```diff
@@ -18,21 +18,26 @@ const router = Router();
 router.get('/:classId/new', async (req, res, next) => {
   try {
     const { classId } = req.params;
-    const { from = '' } = req.query;
+    const { from = '', price = undefined } = req.query;
+    const priceIndex = Number(price) || 0;
```
nwingt: In this case, I would prefer using price_index instead of price, as the latter is more commonly used to refer to the actual price of something
```suggestion
    const { from = '', price_index: priceIndexString = undefined } = req.query;
    const priceIndex = Number(priceIndexString) || 0;
```

src/routes/likernft/book/purchase.ts
```diff
@@ -117,11 +125,11 @@ router.get(
     try {
       const { classId, paymentId } = req.params;
       const doc = await likeNFTBookCollection.doc(classId).collection('transactions').doc(paymentId).get();
+      const docData = doc.data();
       if (!doc.data()) {
```
nwingt:
```suggestion
      const docData = doc.data();
      if (!docData) {
```

src/util/api/likernft/book/index.ts
```diff
@@ -64,15 +71,24 @@ export async function processNFTBookPurchase(
...
+        stock,
+      } = priceInfo;
+      if (stock <= 0) throw new ValidationError('OUT_OF_STOCK');
       const paymentDoc = await t.get(bookRef.collection('transactions').doc(paymentId));
       const paymentData = paymentDoc.data();
       if (!paymentData) throw new ValidationError('PAYMENT_NOT_FOUND');
       if (paymentData.status !== 'new') throw new ValidationError('PAYMENT_ALREADY_CLAIMED');
+      priceInfo.stock -= 1;
+      priceInfo.sold += 1;
+      priceInfo.lastSaleTimestamp = firestore.Timestamp.now();
```
nwingt, line 89: What does `priceInfo` for?

williamchong: Originally `price`(number), `sold`, `stock`, `timestamp` s are all in document root
Now they have to be inside objects in the `prices` array, so here `priceInfo` is the individual `price` object, or `priceInfo` (since `price` can have many meaning...)

williamchong: If you are asking about why assignment without using below, because I am hack-ily updating `prices` array by updating object (they are references)

nwingt (approved):

PR #462, opened 2023-05-18
🥅 Guard listing NFT has been sent case

WeiJun0827:


src/util/api/likernft/listing.ts
```diff
@@ -40,16 +40,32 @@ export function formatListingInfo(info: {
...
   return info;
 }
 
 export async function fetchNFTListingInfoByNFTId(classId: string, nftId: string) {
-  const { data } = await axios.get(`${COSMOS_LCD_INDEXER_ENDPOINT}/likechain/likenft/v1/listings/${classId}/${nftId}`);
-  const info = data.listings[0];
-  return info || null;
+  const [{ data: { listings } }, { data: { owners } }] = await Promise.all([
+    axios.get(`${COSMOS_LCD_INDEXER_ENDPOINT}/likechain/likenft/v1/listings/${classId}/${nftId}`),
+    axios.get(`${COSMOS_LCD_INDEXER_ENDPOINT}/likechain/likenft/v1/owner?class_id=${classId}`),
```
williamchong: I think there is an API that can query the owner of a class id + nft id pair without getting all owners

williamchong (changes requested):

src/util/api/likernft/listing.ts
WeiJun0827: Is it from [chain API](https://github.com/likecoin/likecoin-chain/blob/master/proto/likechain/likenft/v1/query.proto) or [indexer API](https://github.com/likecoin/likecoin-chain-tx-indexer/blob/master/rest/router.go)?

williamchong: from x/nft chain module
https://github.com/likecoin/likecoin-chain/blob/v3.1.1/backport/cosmos-sdk/v0.46.0-rc1/proto/cosmos/nft/v1beta1/query.proto#L19

WeiJun0827: updated, thanks!

williamchong (approved):

PR #467, opened 2023-06-02
🐛 Fix book field name mess up

williamchong:


AuroraHuang22 (approved):

nwingt (commented):
I suggest to rename `stock` to `totalStock` and keep `pStock` as `stock` for less confusing, `sold` as well

PR #468, opened 2023-06-05
👽 Pass claim token to success URL for Stripe sesssion

nwingt:
Requires likecoin/liker-land#1115

### DIFF
--- a/src/routes/likernft/fiat/stripe.ts
+++ b/src/routes/likernft/fiat/stripe.ts
@@ -2,8 +2,9 @@ import { Router } from 'express';
 import bodyParser from 'body-parser';
 import BigNumber from 'bignumber.js';
 import { DeliverTxResponse } from '@cosmjs/stargate';
-import uuidv4 from 'uuid/v4';
 import { TxRaw } from 'cosmjs-types/cosmos/tx/v1beta1/tx';
+import { randomBytes } from 'crypto';
+import uuidv4 from 'uuid/v4';
 
 import Stripe from 'stripe';
 import stripe from '../../../util/stripe';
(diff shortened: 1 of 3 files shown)


src/routes/likernft/fiat/stripe.ts
```diff
@@ -235,6 +237,7 @@ router.post(
           memo,
           iscnPrefix,
           paymentId,
+          claimToken,
```
williamchong: Prefer store it in firestore instead of stripe metadata
Also we should only set claim token when a wallet is not set on purchase
And we would need to void this token if a wallet is found with an email (actually should we keep this email auto claim behavior in the new flow?)

williamchong (changes requested):

src/util/api/likernft/fiat/stripe.ts
```diff
@@ -94,10 +93,10 @@ export async function processStripeFiatNFTPurchase(session, req) {
     wallet = await findWalletWithVerifiedEmail(email);
   }
   const isPendingClaim = !wallet;
-  let claimToken;
+  let claimToken: string | undefined;
   if (isPendingClaim) {
     wallet = LIKER_NFT_PENDING_CLAIM_ADDRESS;
-    claimToken = randomBytes(32).toString('base64url');
+    claimToken = docData.claimToken;
```
williamchong: Not sure if we need to check if `docData.claimToken` exists, need to throw an error maybe

williamchong (approved):

src/util/api/likernft/fiat/stripe.ts
nwingt: Added guard

williamchong (approved):

PR #471, opened 2023-06-08
👽 Use liker.land for claiming NFT Book

nwingt:
Based on #468


williamchong (approved):
Is `NFT_BOOKSTORE_HOSTNAME` used any where else? Can remove if not used

nwingt:
> Is `NFT_BOOKSTORE_HOSTNAME` used any where else? Can remove if not used

Updated

williamchong (approved):

PR #473, opened 2023-06-12
✨ Add book description and locale support

williamchong:
Base on https://github.com/likecoin/likecoin-api-public/pull/472

### DIFF
--- a/src/routes/likernft/book/store.ts
+++ b/src/routes/likernft/book/store.ts
@@ -186,7 +186,9 @@ router.post('/:classId/new', jwtAuth('write:nftbook'), async (req, res, next) =>
       } = p;
       return !(
         typeof name[NFT_BOOK_TEXT_DEFAULT_LOCALE] === 'string'
-        && (description[NFT_BOOK_TEXT_DEFAULT_LOCALE] && typeof description[NFT_BOOK_TEXT_DEFAULT_LOCALE] === 'string'));
+        && Object.values(name).every((n) => typeof n === 'string')
+        && (description[NFT_BOOK_TEXT_DEFAULT_LOCALE] && typeof description[NFT_BOOK_TEXT_DEFAULT_LOCALE] === 'string'))
+        && Object.values(description).every((d) => typeof d === 'string');
     });
     if (invalidNameIndex > -1) {
       throw new ValidationError(`INVALID_NAME_in_${invalidNameIndex}`);
--- a/src/routes/likernft/book/store.ts
+++ b/src/routes/likernft/book/store.ts
@@ -1,6 +1,7 @@
 import { Router } from 'express';
 import {
   MIN_BOOK_PRICE_DECIMAL,
+  NFT_BOOK_TEXT_DEFAULT_LOCALE,
   getNftBookInfo,
   listNftBookInfoByModeratorWallet,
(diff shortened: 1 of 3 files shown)


src/util/ses.ts
```diff
@@ -328,3 +328,41 @@ export function sendNFTBookPendingClaimEmail({
...
+        Value: 'sendNFTBookPendingClaimEmail',
+      },
+    ],
+    Destination: {
+      ToAddresses: emails,
+    },
+    Message: {
+      Subject: {
+        Charset: 'UTF-8',
+        Data: `You have sold an nft for $${amount}`,
```
nwingt:
```suggestion
        Data: `You have sold an NFT for $${amount}`,
```

src/util/api/likernft/book/index.ts
```diff
@@ -110,6 +154,14 @@ export async function processNFTBookPurchase(
...
       });
     }
+    if (notificationEmails.length) {
+      await sendNFTBookSalesEmail({
+        buyerEmail: email,
+        emails: notificationEmails,
+        classId,
+        amount: (amountTotal || 0) / 100,
+      });
+    }
```
nwingt: I think we should also alert the seller when a buyer claimed the NFT, so the seller can take action to sign and send the NFT

src/routes/likernft/book/store.ts
```diff
@@ -160,19 +161,36 @@ router.post('/:classId/new', jwtAuth('write:nftbook'), async (req, res, next) =>
...
       throw new ValidationError(`INVALID_PRICE_in_${invalidPriceIndex}`);
     }
+    const invalidNameIndex = prices.findIndex((p) => {
+      const {
+        name = {},
+        description = {},
+      } = p;
+      return !(
+        typeof name[NFT_BOOK_TEXT_DEFAULT_LOCALE] === 'string'
+        && (description[NFT_BOOK_TEXT_DEFAULT_LOCALE] && typeof description[NFT_BOOK_TEXT_DEFAULT_LOCALE] === 'string'));
```
nwingt: We just check the default language?

src/util/api/likernft/book/index.ts
williamchong: Not sure if we want to alert both on sale and claim, but makes sense

src/routes/likernft/book/store.ts
williamchong: you mean checking string for all locale that exists as a key in `description`?

src/util/ses.ts
williamchong: handle in https://github.com/likecoin/likecoin-api-public/pull/476

src/util/api/likernft/book/index.ts
williamchong: handle in https://github.com/likecoin/likecoin-api-public/pull/476

src/routes/likernft/book/store.ts
nwingt: Yes, or do we want to make sure the default exists?

williamchong: we do want to make sure at least the default exists, but we probably want to make sure any other key that exists is also a string

nwingt (approved):

PR #481, opened 2023-06-26
💬 Use NFT Book name as link text

WeiJun0827:


src/util/api/likernft/book/index.ts
```diff
@@ -98,6 +99,7 @@ export async function processNFTBookPurchase(
   const {
     metadata: {
       classId,
+      iscnPrefix,
```
williamchong, line 102: Please edit https://github.com/likecoin/likecoin-api-public/blob/e8e1787114327cc25116ee8cd90cc192b44dd1fc/src/routes/likernft/book/purchase.ts#L65
to make the metadata field consistent with wnft
now stripe metadata of nft book does not contain `iscnPrefix`

williamchong (changes requested):

williamchong (changes requested):
image

I get undefined name when buying [this](https://rinkeby.liker.land/en/nft/class/likenft1ggkczh9pvsa45f0f2qterymhf6w5y7c2m9v5tmchrn72gzghudtstx4yg2)

src/util/api/likernft/book/index.ts
```diff
@@ -149,12 +151,16 @@ export async function processNFTBookPurchase(
     });
     const { notificationEmails = [] } = listingData;
     const { claimToken } = txData;
-    await stripe.paymentIntents.capture(paymentIntent as string);
+    const [, classData] = await Promise.all([
+      stripe.paymentIntents.capture(paymentIntent as string),
+      getNFTClassDataById(classId),
```
williamchong: Please catch error from `getNFTClassDataById` since we don't want it to stop `stripe.paymentIntents.capture` from running

williamchong (changes requested):

src/util/api/likernft/book/index.ts
WeiJun0827: I see, updated

src/util/ses.ts
```diff
@@ -408,15 +407,20 @@ export function sendNFTBookClaimedEmail({
...
       Body: {
         Html: {
           Charset: 'UTF-8',
           Data: `
-          <p>Dear Liker,</p>
-          <p>${buyerEmail}(${wallet}) claim their NFT book ${classId}${message ? ` with message: ${message}` : ''}
-          <p>Please go to <a href="https://${NFT_BOOKSTORE_HOSTNAME}/nft-book-store/send/${classId}/?payment_id=${paymentId}">NFT book management page</a> to deliver your book NFT.</p>
+          <p>Dear Creator,</p>
+          <br/>
+          <p>Congratulation. A reader has bought your NFT book.</p>
```
williamchong: Should this be ` A reader has claim`? Also if `message` exists should we show `message`?

PR #487, opened 2023-07-05
⚡️ Merge createISCN and changeISCNOwnership msgs in a tx

WeiJun0827:


WeiJun0827:
Tried to see if the `claim` qs was ever used by inspecting the kibana logs, but `claim` wasn't included in the logger, can't seem to determine quickly.
We can keep the feature by the last commit, but is it meaningful for someone to pay and create an ISCN by API wallet?

PR #495, opened 2023-07-21
[TAS-57] ✨ Add plan and subscription related endpoint

williamchong:
- creator apis for reading and creating subscription plans
- stripe apis for creating new subscription, handle subscription update and accessing customer portals
- reader apis for listing current active subscriptions


src/routes/likernft/book/index.ts
```diff
@@ -2,7 +2,7 @@ import { Router } from 'express';
 
 import purchase from './purchase';
 import store from './store';
-import user from './user';
+import user from '../connect';
```
WeiJun0827, line 5: Do we still use `./user.js`?

src/util/api/likernft/subscription/stripe.ts
```diff
@@ -0,0 +1,137 @@
...
+        customerId,
+      },
+      [creatorWallet]: {
+        currentPeriodStart,
+        currentPeriodEnd,
+        subscriptionId,
+        productId,
+        priceId,
+      },
+      timesstamp: Date.now(),
```
WeiJun0827: Should be `timestamp`? May also consider using `FieldValue.serverTimestamp()` here for better readability.

src/util/firebase.ts
```diff
@@ -61,7 +61,7 @@ export const likeNFTSubscriptionTxCollection = getCollectionIfDefined(
   FIRESTORE_NFT_SUBSCRIPTION_TX_ROOT,
 );
 export const likeNFTBookCollection = getCollectionIfDefined(FIRESTORE_LIKER_NFT_BOOK_ROOT);
-export const likeNFTBookUserCollection = getCollectionIfDefined(
+export const likeNFTConnectedUserCollection = getCollectionIfDefined(
   FIRESTORE_LIKER_NFT_BOOK_USER_ROOT,
 );
```
WeiJun0827, line 66: Is `connected` a new term to replace original `book`?

src/util/api/likernft/connect.ts
```diff
@@ -0,0 +1,13 @@
...
+  const userDoc = await likeNFTConnectedUserCollection.doc(wallet).get();
+  const userData = userDoc.data();
+  return userData;
+}
+
+export async function getUserIsStripeConnectReady(wallet: string) {
+  const userData = await getUserStripeConnectInfo(wallet);
+  const { isStripeConnectReady } = userData;
+  return !isStripeConnectReady;
+}
```
WeiJun0827, line 13: Where would we use `getUserIsStripeConnectReady`?

src/routes/likernft/book/index.ts
williamchong: we need to keep it before we change the api path in nft-book-press

src/util/firebase.ts
williamchong: Yes since it is more than just for book now, we would use the same stripe connect for subscription also

src/util/api/likernft/connect.ts
williamchong: We dont use it anywhere now, but seems to be a good helper function for future

src/util/firebase.ts
WeiJun0827: Then should we rename `FIRESTORE_LIKER_NFT_BOOK_USER_ROOT` and its config value too?

PR #505, opened 2023-07-29
[TAS-107] ✨ Support NFT book prices reordering

nwingt:


src/routes/likernft/book/store.ts
```diff
@@ -163,6 +166,58 @@ router.get('/:classId/price/:priceIndex', jwtOptionalAuth('read:nftbook'), async
   }
 });
 
+router.put('/:classId/price/:priceIndex/order', jwtOptionalAuth('write:nftbook'), async (req, res, next) => {
```
williamchong: Auth should not be optional

williamchong (changes requested):

src/routes/likernft/book/store.ts
nwingt: Fixed

PR #506, opened 2023-08-15
[TAS-205] ✨ Support expiry date for claiming WNFT

WeiJun0827:
- Add number parameter `collecatableBefore` for `POST /mint` endpoint
- Set default max timestamp ([ref](https://stackoverflow.com/a/29142324/7978205))
- Store `collecatableBefore` in the Firebase class doc
- Check whether NFT Class is collectable when collecting

### DIFF
--- a/src/routes/likernft/mint.ts
+++ b/src/routes/likernft/mint.ts
@@ -63,9 +63,13 @@ router.post(
         reservedNftCount,
         initialBatch,
         isFree,
+        collectableBefore,
       } = req.body;
+
       if (isFree && initialBatch && initialBatch > -1) throw new ValidationError('CANNOT_SET_BOTH_FREE_AND_INITIAL_BATCH');
       if (!iscnId) throw new ValidationError('MISSING_ISCN_ID');
+      if (collectableBefore && collectableBefore < Date.now()) throw new ValidationError('COLLECTABLE_BEFORE_CANNOT_BE_IN_THE_PAST');
+
       const iscnPrefix = getISCNPrefix(iscnId);
       const iscnPrefixDocName = getISCNPrefixDocName(iscnId);
       const likeNFTDoc = await likeNFTCollection.doc(iscnPrefixDocName).get();
@@ -116,6 +120,7 @@ router.post(
         ...chainMetadata,
         initialBatch,
         isFree,
+        collectableBefore,
         classId,
         totalCount: nfts.length,
         platform,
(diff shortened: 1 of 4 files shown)


nwingt:
I think the param should be named `expiryDate` instead of `collecatableBefore`

WeiJun0827:
> I think the param should be named `expiryDate` instead of `collecatableBefore`

The problem I was thinking about at the time was that the so-called expired action is literally implied, although in our context it is the action of "collecting". Is this a point of concern?

nwingt:
> > I think the param should be named `expiryDate` instead of `collecatableBefore`
> 
> The problem I was thinking about at the time was that the so-called expired action is literally implied, although in our context it is the action of "collecting". Is this a point of concern?

I understand what you meant, how about collectExpiryAt? becoz `before` is a period of time, while `expiryAt` gives the meaning of the exact time

nwingt (approved):

src/util/api/likernft/mint.ts
```diff
@@ -12,6 +12,8 @@ import {
 } from '../../../constant';
 import { sleep } from '../../misc';
 
+const MAX_TIMESTAMP = 8640000000000000;
```
williamchong: Prefer not setting a timestamp value if not necessary, just filter away `undefined` value when setting data in firestore

src/util/api/likernft/mint.ts
```diff
@@ -86,6 +89,7 @@ export async function writeMintedNFTInfo(iscnPrefix, classData, nfts) {
     soldCount: 0,
     creatorWallet: sellerWallet,
     timestamp,
+    collectExpiryAt,
```
williamchong: I think we can put this in `iscnRef`, since most logic (batch, remaining count, etc) happen in `iscnRef` anyway. `iscnRef` and `nftClass` are now strictly 1 to 1, if we want to change that we would have to move these existing logic anyway

src/routes/likernft/mint.ts
```diff
@@ -63,9 +63,13 @@ router.post(
         reservedNftCount,
         initialBatch,
         isFree,
+        collectExpiryAt,
```
williamchong, line 66: `collectableBefore` or `collectExpiryAt`?

WeiJun0827: From the prior discussion with Tat, `collectExpiryAt` was suggested. But further discussion is welcome 🌝
I will update nft-book-press PR accordingly if this PR is finalized.

src/util/api/likernft/mint.ts
WeiJun0827: OK, I see!

williamchong (approved):

PR #508, opened 2023-08-17
[TAS-231] ✨ Allow update NFT Book prices

WeiJun0827:


src/routes/likernft/book/store.ts
```diff
@@ -293,10 +264,12 @@ router.post('/:classId/settings', jwtAuth('write:nftbook'), async (req, res, nex
   try {
     const { classId } = req.params;
     const {
+      prices = [],
```
williamchong: If `prices` is not set in body, then all existing `prices` would be wiped if we have a default array
Should just leave it as `undefined`, skip checking if prices if not set, and don't update if prices is not set

williamchong (changes requested):

src/routes/likernft/book/store.ts
WeiJun0827: OK, leave it as `undefined`, and let `updateNftBookSettings` check the payload.
https://github.com/likecoin/likecoin-api-public/blob/master/src/util/api/likernft/book/index.ts#L78

williamchong (approved):

PR #512, opened 2023-08-22
🐛 Guard nextNewNFTId if NFT has expired

WeiJun0827:


WeiJun0827:
Hmmm... actually the expiry time has been guarded in `getPriceInfo` if someone tries to purchase an expired NFT. And the `nextNewNFTId` was intact before.
https://github.com/likecoin/likecoin-api-public/blob/master/src/util/api/likernft/purchase.ts#L166-L168
Should we set `nextNewNFTId` so that [line 160](https://github.com/likecoin/likecoin-api-public/blob/master/src/util/api/likernft/purchase.ts#L160) will throw a `SELLING_NFT_DOC_NOT_FOUND` error?

WeiJun0827:
Another off-the-topic concern is that I found `getLatestNFTPriceAndInfo` is called twice for each NFT in `POST /purchase` API.
It is [called](https://github.com/likecoin/likecoin-api-public/blob/master/src/routes/likernft/purchase.ts#L72) to calculate and check the granted amount the first time.
And [called again](https://github.com/likecoin/likecoin-api-public/blob/master/src/util/api/likernft/purchase.ts#L159) to process the purchase in `processNFTPurchase` -> `getPriceInfo`.

Should we move the check of the first step into the second one to prevent race condition/inconsistency?

src/util/api/likernft/purchase.ts
```diff
@@ -137,7 +137,7 @@ export async function getLatestNFTPriceAndInfo(
     price = currentPrice;
     // This NFT ID represents a possible NFT of that NFT Class for purchasing only,
     // another fresh one might be used on purchase instead
-    nextNewNFTId = newNftDocData.id;
+    if (!collectExpiryAt || collectExpiryAt < Date.now()) nextNewNFTId = newNftDocData.id;
     isProcessing = newNftDocData.isProcessing;
```
williamchong, line 141: maybe we should not set anything if `collectExpiryAt` is expired, act like there is no `newNftDocData`

williamchong (changes requested):

williamchong:
> Another off-the-topic concern is that I found `getLatestNFTPriceAndInfo` is called twice for each NFT in `POST /purchase` API. It is [called](https://github.com/likecoin/likecoin-api-public/blob/master/src/routes/likernft/purchase.ts#L72) to calculate and check the granted amount the first time. And [called again](https://github.com/likecoin/likecoin-api-public/blob/master/src/util/api/likernft/purchase.ts#L159) to process the purchase in `processNFTPurchase` -> `getPriceInfo`.
> 
> Should we move the check of the first step into the second one to prevent race condition/inconsistency?

The original idea was to check the tx info and the price matches before going into more expensive db transaction, we should no rely on  any information provided by first check in the later sections.

src/util/api/likernft/purchase.ts
WeiJun0827: OK, I should close this PR

PR #515, opened 2023-08-24
🎨 Store `collectExpiryAt` in Firebase DateTime format

WeiJun0827:


WeiJun0827:
@williamchong do you know why the first commit code doesn't work?

williamchong:
> @williamchong do you know why the first commit code doesn't work?

How does it "not work"?

WeiJun0827:
> How does it "not work"?

It runs OK at my local, but it can't pass the `npm run test`.

williamchong:
> > How does it "not work"?
> 
> It runs OK at my local, but it can't pass the `npm run test`.

Our firestore stub used for test doesn't have Timestamp stub and exported i guess

WeiJun0827:
> Our firestore stub used for test doesn't have Timestamp stub and exported i guess

Oh, it makes sense, thanks!

williamchong (approved):

PR #518, opened 2023-09-08
🥅 Can only purchase one NFT per class

WeiJun0827:


src/middleware/likernft.ts
```diff
@@ -36,7 +36,7 @@ export const fetchISCNPrefixes = async (req, res, next) => {
   try {
     const { class_id: classId } = req.query;
     if (!classId) throw new ValidationError('MISSING_ISCN_OR_CLASS_ID');
-    const classIds = Array.isArray(classId) ? classId : [classId];
+    const classIds = Array.isArray(classId) ? Array.from(new Set(classId)) : [classId];
```
williamchong: prefer throwing error instead of silent remove

williamchong (changes requested):

williamchong (approved):

nwingt (approved):

PR #521, opened 2023-09-11
[TAS-374] ✨ Support purchase multiple wNFT by Stripe

WeiJun0827:


src/util/api/likernft/index.ts
```diff
@@ -32,15 +32,6 @@ export async function getISCNPrefixByClassId(classId) {
   return decodeURIComponent(doc.id);
 }
 
-export async function getISCNPrefixByClassIdFromChain(classId) {
```
williamchong: Why remove from chain functions?

src/routes/likernft/fiat/stripe.ts
```diff
@@ -141,81 +114,67 @@ router.get(
...
   async (req, res, next) => {
     try {
-      const classId = req.query.class_id as string;
       const { wallet } = req.query;
       if (!(wallet || LIKER_NFT_PENDING_CLAIM_ADDRESS) && !isValidLikeAddress(wallet)) throw new ValidationError('INVALID_WALLET');
-      const { iscnPrefix } = res.locals;
-      const promises = [getNFTClassDataById(classId)];
       const {
-        nftId = '', seller = '', memo, email,
+        collects, memo, email,
```
williamchong: Please name it `collectInfoList` or `collectNftInfoList` or similar names

src/util/api/likernft/fiat/stripe.ts
```diff
@@ -276,3 +281,84 @@ export async function processStripeFiatNFTPurchase(session, req) {
...
+    const { nftId, seller } = firstListing;
+    result.nftId = nftId;
+    result.seller = seller;
+  }
+  return result;
+}
+
+export async function getNFTPurchaseInfo({
+  classId,
+  nftId = '',
```
williamchong: Is nftId a required param here?

williamchong (changes requested):
Need to also handle `processFiatNFTPurchase`

src/util/api/likernft/index.ts
WeiJun0827: No function is using it, can still keep it

src/util/api/likernft/fiat/stripe.ts
WeiJun0827: It's optional, only for listing class

src/routes/likernft/fiat/stripe.ts
```diff
@@ -141,81 +99,63 @@ router.get(
...
+        name = `${name} + ${classMetadataList.length - 1} more`;
+      }
+
       const claimToken = randomBytes(32).toString('base64url');
+
+      const iscnAndClassIdLog = {};
+      classIds.forEach((classId, i) => {
+        iscnAndClassIdLog[`iscnPrefix${i + 1}`] = iscnPrefixes[i];
+        iscnAndClassIdLog[`classId${i + 1}`] = classId;
+      });
```
williamchong, line 130: Interesting and smart way to populate stripe metadata

src/routes/likernft/fiat/stripe.ts
```diff
@@ -141,81 +99,63 @@ router.get(
...
+          classId: classIds[0],
           paymentId,
           token: claimToken,
           wallet: wallet as string,
         }),
         client_reference_id: wallet as string || undefined,
         customer_email: email,
-        cancel_url: getLikerLandNFTClassPageURL({ classId }),
+        cancel_url: getLikerLandNFTClassPageURL({ classId: classIds[0] }),
         line_items: [
```
williamchong: Actually we can list each nft as a single line item, but that would make price calculation very hard in LIKE case

williamchong (approved):

src/routes/likernft/fiat/stripe.ts
WeiJun0827: Or should we split the `LIKER_NFT_FIAT_FEE_USD` as an additional line item?

williamchong (approved):

williamchong (approved):

PR #522, opened 2023-09-18
[TAS-402] 👔 Change wNFT base currency to USD

WeiJun0827:


src/util/api/likernft/fiat/stripe.ts
```diff
@@ -315,14 +314,14 @@ export function formatLineItem(classMetadata, fiatPrice) {
...
-export function getTxFeeLineItem() {
+export function formatTxFeeLineItem(txFee) {
   return {
     price_data: {
       currency: 'USD',
       product_data: {
         name: 'Transaction Fee',
       },
-      unit_amount: Number(new BigNumber(LIKER_NFT_FIAT_FEE_USD).shiftedBy(2).toFixed(0)),
+      unit_amount: Number(new BigNumber(txFee).shiftedBy(2).toFixed(0)),
```
williamchong: Should set round up?

src/routes/likernft/purchase.ts
```diff
@@ -35,10 +34,9 @@ router.get(
         if (isFree && wallet) {
           canFreeCollect = !(await checkFreeMintExists(wallet as string, classId));
         }
-        const gasFee = isFree ? 0 : getGasPrice();
+        const gasFee = 0;
```
williamchong: Please add comment to explain hardcode

src/util/api/likernft/fiat/index.ts
```diff
@@ -54,40 +53,37 @@ export async function getPurchaseInfoList(iscnPrefixes, classIds) {
...
-    .reduce((acc, { LIKEPrice }) => acc.plus(LIKEPrice), new BigNumber(0)).toFixed(9));
-  const fiatPrices = purchaseInfoList.map(
-    ({ LIKEPrice }) => new BigNumber(LIKEPrice)
-      .multipliedBy(rate)
-      .multipliedBy(1 + buffer)
-      .toFixed(2, BigNumber.ROUND_CEIL),
-  );
-  let totalFiatBigNum = fiatPrices.reduce((acc, p) => acc.plus(p), new BigNumber(0));
-  if (totalFiatBigNum.gt(0)) totalFiatBigNum = totalFiatBigNum.plus(LIKER_NFT_FIAT_FEE_USD);
+    .reduce((acc, { price }) => acc.plus(price), new BigNumber(0)).dividedBy(rate).toFixed(0));
```
williamchong: Please round up

src/util/api/likernft/fiat/stripe.ts
WeiJun0827: hmmm `txFee` is already a round-up value?

williamchong: if we are calling `toFixed(0)` anyway might as well just define round up mode

src/util/api/likernft/purchase.ts
```diff
@@ -632,9 +669,11 @@ export async function processNFTPurchase({
     return _priceInfoList;
   });
 
-  const dbNFTTotalPrice = priceInfoList.reduce((acc, d) => acc + d.price, 0);
+  const LIKEPrices = priceInfoList.map((info) => Number(new BigNumber(info.price)
+    .dividedBy(LIKEPriceInUSD).toFixed(0, BigNumber.ROUND_UP)));
+  const totalLIKEPrice = LIKEPrices.reduce((acc, p) => acc + p, 0);
 
-  if (dbNFTTotalPrice && dbNFTTotalPrice > grantedAmount) {
+  if (totalLIKEPrice && totalLIKEPrice > grantedAmount) {
```
williamchong, line 685: Can we leave some buffer when checking/report price? We should report price which over estimate LIKE price, then check the price needed here with actual LIKE price
This is to avoid small changes in LIKE price would make the whole tx fail

src/util/api/likernft/purchase.ts
```diff
@@ -40,6 +44,27 @@ import {
...
+  if (hasCache) {
+    price = cachedPrice;
+  } else {
+    price = await axios.get(COINGECKO_PRICE_URL)
+      .then((r) => {
+        const p = r.data.market_data.current_price[CURRENCY];
+        priceCache.set(CURRENCY, p);
+        return p;
+      })
+      .catch(() => cachedPrice || LIKER_NFT_FIAT_MIN_RATIO);
```
williamchong: We can use async await here without .then and .catch?

src/util/api/likernft/purchase.ts
WeiJun0827: Is it OK to add the buffer in the labeled LIKE price & granted amount, then deduct the actual LIKE price later?
(See commit https://github.com/likecoin/likecoin-api-public/pull/522/commits/797fd55d09edf1514f962a85ce6eb2cd42b1f314)

Currently, if an NFT of $1 USD is worth $800 LIKE, the shown LIKE price will be $880 LIKE, and request a grant of $880 LIKE. So either LIKE price goes up or down less than 10%, the tx still success, and the LIKE price won't be underestimated if LIKE price goes up. (if we instead use the actual price of $800 LIKE to request a grant then check with buffer later)

williamchong: yes current one would be fine

src/util/api/likernft/purchase.ts
```diff
@@ -634,9 +673,20 @@ export async function processNFTPurchase({
     return _priceInfoList;
   });
 
-  const dbNFTTotalPrice = priceInfoList.reduce((acc, d) => acc + d.price, 0);
+  const LIKEPrices = priceInfoList.map((info) => {
+    let bigNum = new BigNumber(info.price);
+    if (shouldDeductStripeTxFee) {
```
williamchong: Can we handle stripe fee in `processFiatNFTPurchase` or `processStripeFiatNFTPurchase` ? Since this is totally a stripe or fiat only thing

williamchong (changes requested):

src/util/api/likernft/purchase.ts
WeiJun0827: Yes, that would be better. But how could we deduct the tx fee before knowing the NFT price? Since the price is fetched inside of `processNFTPurchase`.

src/util/api/likernft/purchase.ts
```diff
@@ -648,6 +648,7 @@ export async function processNFTPurchase({
   grantTxHash = '',
   granterMemo = '',
   retryTimes = 0,
+  txFeeReducer = (p: BigNumber) => p,
```
williamchong: Confusing name, the thing being reduce is revenue/profit not txFee, how about priceModifier or sth like that?

src/util/api/likernft/purchase.ts
```diff
@@ -672,8 +673,10 @@ export async function processNFTPurchase({
     return _priceInfoList;
   });
 
-  const LIKEPrices = priceInfoList.map((info) => Number(new BigNumber(info.price)
-    .dividedBy(LIKEPriceInUSD).toFixed(0, BigNumber.ROUND_UP)));
+  const LIKEPrices = priceInfoList.map((info) => {
+    const reducedFiatPrice = txFeeReducer(new BigNumber(info.price));
```
williamchong: better check txFeeReducer is defined and is a function

williamchong (approved):

williamchong (approved):

PR #528, opened 2023-09-29
✨ Reactivate verification email

williamchong:


src/util/api/users/register.ts
```diff
@@ -76,7 +77,7 @@ export async function suggestAvailableUserName(username = '') {
 export async function handleUserRegistration({
   payload,
   req,
-  // res,
+  res,
```
WeiJun0827, line 80: Should we update the caller params here?
https://github.com/williamchong/likecoin-api-public/blob/7277ac3c2ddb324cd55c8141708a18b233b73a12/src/routes/users/registerLogin.ts#L195-L203

williamchong: yes indeed

WeiJun0827 (approved):

PR #530, opened 2023-10-03
[TAS-488] ✨ Deduct the amount of Stripe tx fee for NFT checked-out with LIKE

WeiJun0827:


config/config.js
```diff
@@ -53,6 +53,7 @@ config.LIKER_NFT_MINT_MIN_BATCH = 7; // limit for USD base
 config.LIKER_NFT_LIKE_TO_USD_CONVERT_RATIO = 1024;
 config.LIKER_NFT_MIN_USD_PRICE = 1;
 config.LIKER_NFT_GAS_FEE = '200000';
+config.LIKER_NFT_PAY_BY_LIKE_PRICE_OFF = 0.4;
```
nwingt:
```suggestion
config.LIKER_NFT_LIKE_PRICE_DISCOUNT_IN_USD = 0.4;
```

src/routes/likernft/purchase.ts
```diff
@@ -93,6 +98,9 @@ router.post(
         likeWallet = wallet as string;
         grantedAmount = 0;
       }
+      const priceModifier = (p: BigNumber) => p
+        .minus(LIKER_NFT_TX_FEE_DISCOUNT_IN_USD)
+        .multipliedBy(1 - LIKER_NFT_LIKE_PRICE_DISCOUNT_RATIO);
```
nwingt: can we refactor this function to util and reuse in `calculatePayment()`

src/util/api/likernft/fiat/index.ts
```diff
@@ -37,15 +39,28 @@ export async function getPurchaseInfoList(iscnPrefixes, classIds) {
...
+      const discountedPrice = new BigNumber(price)
+        .minus(LIKER_NFT_TX_FEE_DISCOUNT_IN_USD)
+        .multipliedBy(1 - LIKER_NFT_LIKE_PRICE_DISCOUNT_RATIO);
+      return discountedPrice.isLessThanOrEqualTo(0)
+        ? acc
+        : acc.plus(discountedPrice);
+    }, new BigNumber(0))
+    .dividedBy(rate)
+    .multipliedBy(1 + buffer)
+    .toFixed(0, BigNumber.ROUND_UP));
```
nwingt: How about?
```js
const priceReducer = (acc, { price }) => {
  const discountedPrice = new BigNumber(price)
    .minus(LIKER_NFT_TX_FEE_DISCOUNT_IN_USD)
    .multipliedBy(1 - LIKER_NFT_LIKE_PRICE_DISCOUNT_RATIO);
  return discountedPrice.isLessThanOrEqualTo(0)
    ? acc
    : acc.plus(discountedPrice);
};
const priceReducerWithoutDiscount = (acc, { price }) => acc.plus(price);

const calculateTotalLIKEPrice = reducer => Number(
  purchaseInfoList
    .reduce(reducer, new BigNumber(0))
    .dividedBy(rate)
    .multipliedBy(1 + buffer)
    .toFixed(0, BigNumber.ROUND_UP)
);

const totalLIKEPrice = calculateTotalLIKEPrice(priceReducer);
const totalLIKEPriceNoDiscount = calculateTotalLIKEPrice(priceReducerWithoutDiscount);
```

WeiJun0827: I extracted the modifier and used the same reducer for both cases, to guard exceptional/unwanted cases.

config/config.js
```diff
@@ -53,6 +53,8 @@ config.LIKER_NFT_MINT_MIN_BATCH = 7; // limit for USD base
 config.LIKER_NFT_LIKE_TO_USD_CONVERT_RATIO = 1024;
 config.LIKER_NFT_MIN_USD_PRICE = 1;
 config.LIKER_NFT_GAS_FEE = '200000';
+config.LIKER_NFT_TX_FEE_DISCOUNT_IN_USD = 0.5;
+config.LIKER_NFT_LIKE_PRICE_DISCOUNT_RATIO = 0.05;
```
williamchong: Should not hardcode LIKE discount as conflict, we should just calculate fiat price - stripe fee and use that as discounted price

williamchong (changes requested):

src/util/api/likernft/fiat/index.ts
```diff
@@ -37,15 +40,29 @@ export async function getPurchaseInfoList(iscnPrefixes, classIds) {
...
+      .dividedBy(rate)
+      .multipliedBy(1 + buffer)
+      .toFixed(0, BigNumber.ROUND_UP),
+  );
+  const priceModifierWithoutDiscount = (price: BigNumber) => price;
+
+  const totalLIKEPrice = calculateTotalLIKEPrice(priceModifierForCheckoutWithLIKE);
+  const totalLIKEPriceNoDiscount = calculateTotalLIKEPrice(priceModifierWithoutDiscount);
   const totalFiatBigNum = purchaseInfoList
     .reduce((acc, { price }) => acc.plus(price), new BigNumber(0));
```
nwingt: I wonder why `purchaseInfoList` is `{ price }[]` here but `price[]` in above

WeiJun0827: Indeed, should add a `map` before calling reducer.

src/routes/likernft/payment.ts
```diff
@@ -11,8 +11,13 @@ router.get(
     try {
       const { iscnPrefixes, classIds } = res.locals;
       const purchaseInfoList = await getPurchaseInfoList(iscnPrefixes, classIds);
-      const { totalLIKEPrice, totalFiatPriceString } = await calculatePayment(purchaseInfoList);
+      const {
+        totalLIKEPriceNoDiscount,
```
williamchong: How about prediscount instead of nodiscount

src/util/api/likernft/purchase.ts
```diff
@@ -639,6 +641,21 @@ async function updateDocsForMissingSoldNFT(t, {
   });
 }
 
+// NOTE: Stripe fee per order is 5.4% + 30 cents
+export function rewardModifierForCheckoutWithUSD(priceInUSD: BigNumber, nftCountOfOrder: number) {
+  return priceInUSD
+    // all NFT share flat fee
+    .minus(new BigNumber(LIKER_NFT_STRIPE_FEE_USD_INTERCEPT).dividedBy(nftCountOfOrder))
```
williamchong: For consistency we should just minus the flat price for all nft?

williamchong (approved):

PR #531, opened 2023-10-03
[TAS-491] ✨ Add list free NFT API

williamchong:


WeiJun0827 (commented):
Should we also filter those that have expired?

WeiJun0827 (approved):

PR #532, opened 2023-10-04
[TAS-490] ✨ Add purchase free NFT book API

williamchong:


src/routes/likernft/book/purchase.ts
```diff
@@ -202,6 +219,97 @@ router.get('/:classId/new', async (req, res, next) => {
...
+        isPendingClaim: false,
+        claimToken,
+        classId,
+        priceInDecimal,
+        price: priceInDecimal / 100,
+        priceName,
+        priceIndex,
+        from,
+        status: 'new',
+        timestamp: FieldValue.serverTimestamp(),
```
nwingt, line 276: Simliar to https://github.com/williamchong/likecoin-api-public/blob/29d966026312de1d57fcefa2e236f0f767cddbf1/src/routes/likernft/book/purchase.ts#L174-L188

Can be refactorized

src/util/liker-land.ts
```diff
@@ -7,18 +7,22 @@ export const getLikerLandNFTClassPageURL = ({ classId, language }: { classId: st
...
   classId: string;
   paymentId: string;
+  free?: boolean,
   token: string;
   type?: string;
   language?: string;
   redirect?: boolean;
-}) => getLikerLandURL(`/nft/claim?class_id=${classId}&payment_id=${paymentId}&claiming_token=${token}${redirect ? '&redirect=1' : ''}${type ? `&type=${type}` : ''}`, { language });
+  priceIndex?: number;
+}) => getLikerLandURL(`/nft/claim?class_id=${classId}&payment_id=${paymentId}&claiming_token=${token}${redirect ? '&redirect=1' : ''}${type ? `&type=${type}` : ''}${free ? '&free=1' : ''}${priceIndex !== undefined ? `&price_index=${priceIndex}` : ''}`, { language });
```
nwingt, line 25: So painful to read
```suggestion
}) => {
  const qsPayload = {
    class_id: classId,
    payment_id: paymentId,
    claiming_token: token,
  };
  if (redirect) {
    qsPayload.redirect = '1';
  }
  if (type) {
    qsPayload.type = type;
  }
  if (free) {
    qsPayload.free = '1';
  }
  if (priceIndex !== undefined) {
    qsPayload.price_index = priceIndex;
  }
  const qs = Object.entries(qsPayload).map(([key, value]) => `${key}=${value}`).join('&');
  return getLikerLandURL(`/nft/claim?${qs}`, { language });
};
```

src/util/api/likernft/book/index.ts
```diff
@@ -317,9 +356,9 @@ export function validatePrice(price: any) {
     name = {},
     description = {},
   } = price;
-  if (!(priceInDecimal > 0
+  if (!(priceInDecimal >= 0
     && (typeof priceInDecimal === 'number')
-    && priceInDecimal >= MIN_BOOK_PRICE_DECIMAL)) {
+    && (priceInDecimal === 0 || priceInDecimal >= MIN_BOOK_PRICE_DECIMAL))) {
```
nwingt, line 361:
```suggestion
  if (!(
    priceInDecimal >= 0
    && typeof priceInDecimal === 'number'
    && (priceInDecimal === 0 || priceInDecimal >= MIN_BOOK_PRICE_DECIMAL)
  )) {
```

src/util/liker-land.ts
WeiJun0827: Or is it possible to format with `URLSearchParams`?

WeiJun0827 (commented):
Reviewed without additional comments

src/util/liker-land.ts
williamchong: Actually yes, but actually I think we all use `querystring` else where, actually should use `querystring` to keep things consistent..

PR #537, opened 2023-10-06
♻️ Refactor free book and payment tx creation functions

williamchong:
Follow up on https://github.com/likecoin/likecoin-api-public/pull/532

### DIFF
--- a/src/util/api/likernft/book/index.ts
+++ b/src/util/api/likernft/book/index.ts
@@ -356,9 +356,11 @@ export function validatePrice(price: any) {
     name = {},
     description = {},
   } = price;
-  if (!(priceInDecimal >= 0
-    && (typeof priceInDecimal === 'number')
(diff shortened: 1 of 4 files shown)


WeiJun0827 (approved):

src/util/api/likernft/book/index.ts
```diff
@@ -356,9 +356,11 @@ export function validatePrice(price: any) {
     name = {},
     description = {},
   } = price;
-  if (!(priceInDecimal >= 0
-    && (typeof priceInDecimal === 'number')
-    && (priceInDecimal === 0 || priceInDecimal >= MIN_BOOK_PRICE_DECIMAL))) {
+  if (!(
+    priceInDecimal >= 0
+    && typeof priceInDecimal === 'number'
```
nwingt: Should check type first?

PR #542, opened 2023-11-01
[TAS-566] ✨ Add `POST /:classId/price` API for adding new NFT Book edition

WeiJun0827:


src/routes/likernft/book/store.ts
```diff
@@ -189,6 +189,28 @@ router.get('/:classId/price/:priceIndex', jwtOptionalAuth('read:nftbook'), async
...
+    const bookInfo = await getNftBookInfo(classId);
+    if (!bookInfo) throw new ValidationError('BOOK_NOT_FOUND', 404);
+
+    const { prices = [] } = bookInfo;
+    const newPriceInfo = formatPriceInfo(price);
+    prices.push(newPriceInfo);
+
+    await updateNftBookInfo(classId, { prices });
+    res.json({
+      index: prices.length - 1,
```
williamchong, line 215: Should the input contains `index`, then we compare the index does not already exists, so that we know that user is not adding duplicated price?

WeiJun0827: Should the use case be adding a new price at the end of prices, or adding with specific price index (where user should know the prices length first)?

src/routes/likernft/book/store.ts
```diff
@@ -189,6 +189,35 @@ router.get('/:classId/price/:priceIndex', jwtOptionalAuth('read:nftbook'), async
   }
 });
 
+router.post('/:classId/price', jwtAuth('write:nftbook'), async (req, res, next) => {
+  try {
+    const { classId } = req.params;
+    const { price, priceIndex } = req.body;
```
williamchong: Should `priceIndex` be in path for consistency?

src/routes/likernft/book/store.ts
williamchong: end of prices

williamchong (approved):

PR #544, opened 2023-11-01
[TAS-527] ✨ Support purchasing NFT Book by LIKE

WeiJun0827:


WeiJun0827:
Example link:
https://api.rinkeby.like.co/likernft/book/payment/price?class_id=likenft14fhmat075g4wvmxxlj8shr4mryt8czd5q493mhz99luudxglra4ql69r00&price_index=0

src/routes/likernft/book/index.ts
```diff
@@ -1,11 +1,13 @@
...
 
+import payment from './payment';
 import purchase from './purchase';
 import store from './store';
 import user from './user';
 
 const router = Router();
 
+router.use('/book/payment', payment);
 router.use('/book/purchase', purchase);
```
williamchong: can the `payment` route be under `purchase` also? both seems similar in meaning

src/routes/likernft/book/payment.ts
```diff
@@ -0,0 +1,41 @@
...
+      }
+
+      const { priceInDecimal } = prices[priceIndex];
+      const price = priceInDecimal / 100;
+
+      const {
+        totalLIKEPricePrediscount,
+        totalLIKEPrice,
+        totalFiatPriceString,
+      } = await calculatePayment([price]);
```
williamchong: Please include gas price also

src/routes/likernft/book/purchase.ts
```diff
@@ -360,6 +362,79 @@ router.post(
   },
 );
 
+router.post(
+  '/:classId/new/like',
+  async (req, res, next) => {
+    try {
+      const { classId } = req.params;
+      const { price_index: priceIndexString, tx_hash: grantTxHash } = req.query;
+      const priceIndex = Number(priceIndexString);
```
williamchong: Consider accepting `email` as input as body
And since we are accepting a body, `tx_hash` might as well be in body too

src/util/api/likernft/book/index.ts
```diff
@@ -364,11 +391,12 @@ export async function claimNFTBook(
     const {
       claimToken,
       status,
+      type,
     } = docData;
     if (token !== claimToken) {
       throw new ValidationError('INVALID_CLAIM_TOKEN', 403);
     }
-    if (status !== 'paid') {
+    if (status !== 'paid' && type !== 'LIKE') {
```
williamchong: Maybe check for an expected status also if `type === 'LIKE'`

src/routes/likernft/book/index.ts
WeiJun0827: Well, yes, sort of... we  have `GET /likernft/payment/price` to query LIKE price of W.NFT.
So the `GET /likernft/book/payment/price` is named to keep the symmetric. 
But `GET /likernft/book/:classId/price` would probably be better I think.

src/util/api/likernft/book/index.ts
WeiJun0827: Seems I shouldn't set this check, just leave it unchanged

src/routes/likernft/book/purchase.ts
```diff
@@ -360,6 +396,90 @@ router.post(
...
+      const checkResult = await checkTxGrantAndAmount(grantTxHash, LIKEPrice);
+      if (!checkResult) throw new ValidationError('SEND_GRANT_NOT_FOUND');
+      const {
+        granter: granterWallet,
+        memo: message,
+      } = checkResult;
+
+      const execTxHash = await execAuthz(granterWallet, ownerWallet, LIKEPrice);
+      const paymentId = execTxHash;
+      await createNewNFTBookPayment(classId, paymentId, {
```
williamchong: I think to align with stripe payment flow, the order should be `createNewNFTBookPayment()` => `execAuthz()` => `processNFTBookPurchase()`, since new payment is created before actual payment.
In this case we can also just use ```const paymentId = uuidv4();
      const claimToken = crypto.randomBytes(32).toString('hex');``` instead of hash

williamchong (changes requested):

src/routes/likernft/book/purchase.ts
WeiJun0827: Should we store `execTxHash` somewhere?

williamchong: Yes store it inside the payment object

src/routes/likernft/book/purchase.ts
```diff
@@ -360,6 +396,90 @@ router.post(
...
 
+router.post(
+  '/:classId/new/like',
+  async (req, res, next) => {
+    try {
+      const { classId } = req.params;
+      const { price_index: priceIndexString } = req.query;
+      const {
+        email,
+        txHash: grantTxHash,
```
williamchong, line 434: Should handle `from` parameter because we need this for calculating commission of channels

Ref: https://github.com/likecoin/likecoin-api-public/blob/master/src/routes/likernft/book/purchase.ts#L73

src/routes/likernft/book/purchase.ts
```diff
@@ -360,6 +396,90 @@ router.post(
...
+      const { totalLIKEPrice: LIKEPrice } = await calculatePayment([price]);
+
+      const checkResult = await checkTxGrantAndAmount(grantTxHash, LIKEPrice);
+      if (!checkResult) throw new ValidationError('SEND_GRANT_NOT_FOUND');
+      const {
+        granter: granterWallet,
+        memo: message,
+      } = checkResult;
+
+      const execTxHash = await execAuthz(granterWallet, ownerWallet, LIKEPrice);
```
williamchong: Actually if `from` === `liker_land`, we would want to get 30%. 
Also we collect 5% for commission for all purchase, no matter channel.
Ref: https://github.com/likecoin/likecoin-api-public/blob/master/src/routes/likernft/book/purchase.ts#L148

williamchong (changes requested):

src/routes/likernft/book/purchase.ts
WeiJun0827: Updated, tho still wondering if `claimToken` is meaningful here?

williamchong: No unless we want to "resend" a link in the future, which might happen for gift feature (not sure yet)

src/util/api/likernft/book/index.ts
```diff
@@ -159,6 +175,52 @@ export async function listNftBookInfoByModeratorWallet(moderatorWallet: string)
...
+      .toFixed(0, BigNumber.ROUND_CEIL)
+    : '0';
+  const totalFeeAmount = new BigNumber(gasFeeAmount)
+    .plus(likerLandFeeAmount)
+    .plus(likerLandCommission)
+    .toFixed();
+  const profitAmount = distributedAmountBigNum
+    .minus(likerLandFeeAmount)
+    .minus(likerLandCommission)
+    .toFixed();
```
williamchong, line 204: Fee and commission needs to be sent to `LIKER_NFT_FEE_ADDRESS` instead of `LIKER_NFT_TARGET_ADDRESS`

williamchong (changes requested):

williamchong (approved):

PR #546, opened 2023-11-03
🔧 Split arweave v1 and v2 size config

williamchong:


src/routes/arweave/index.ts
```diff
@@ -1,7 +1,8 @@
 import { Router } from 'express';
 import multer from 'multer';
 import {
-  ARWEAVE_MAX_SIZE,
+  ARWEAVE_MAX_SIZE_V2,
```
AuroraHuang22: Is this being used?

williamchong: No, strange that unused var is not an error..

AuroraHuang22 (approved):

WeiJun0827 (approved):

PR #553, opened 2023-11-17
📈 Accept ga client id as stripe metadata

williamchong:
Ref: https://stripe.com/docs/payments/checkout/analyze-conversion-funnel#link-client-and-server-side-events


nwingt (approved):

williamchong:
@nwingt wrongly put the data into product metadata instead of payment metadata, please review again

nwingt (approved):

WeiJun0827 (approved):

PR #570, opened 2023-11-30
💄 Show from string in stripe checkout ui

williamchong:


src/routes/likernft/book/purchase.ts
```diff
@@ -110,6 +110,8 @@ router.get('/:classId/new', async (req, res, next) => {
     } else if (priceDescription) {
       description = `${description} - ${priceDescription}`;
     }
+
+    if (from) description = `(${from}) ${description}`;
```
nwingt: I think square bracket is better
```suggestion
    if (from) description = `[${from}] ${description}`;
```

nwingt (approved):

PR #573, opened 2023-12-07
[TAS-801] ✨ Send notification email and Slack message for NFT Book listing and sales

WeiJun0827:


src/util/slack.ts
```diff
@@ -0,0 +1,146 @@
...
+  paymentId,
+  method,
+}) {
+  if (!NFT_BOOK_SALES_NOTIFICATION_WEBHOOK) return;
+  try {
+    await axios.post(NFT_BOOK_SALES_NOTIFICATION_WEBHOOK, {
+      classId,
+      className,
+      paymentId,
+      method,
```
nwingt, line 173: - Please also include price and currency
- name of the NFT Class(Link to liker.land)

nwingt (commented):
Can we use block format for Slack message like we did in `sendStripeFiatPurchaseSlackNotification`

WeiJun0827:
> Can we use block format for Slack message like we did in `sendStripeFiatPurchaseSlackNotification`

Block format would be easier to read.
But AFAIK, only Slack App supports it, Slack Workflow supports non-nested JSON only.

Should use Slack App if block format is required.

ref: https://slack.com/intl/zh-tw/help/articles/360041352714-%E5%BB%BA%E7%AB%8B%E9%80%8F%E9%81%8E-Webhook-%E5%95%9F%E5%8B%95%E7%9A%84%E5%B7%A5%E4%BD%9C%E6%B5%81%E7%A8%8B#:~:text=%E8%AB%8B%E7%95%99%E6%84%8F%EF%BC%8C%E7%9B%AE%E5%89%8D%E7%84%A1%E6%B3%95%E5%9C%A8%E5%B7%A5%E4%BD%9C%E6%B5%81%E7%A8%8B%E8%AE%8A%E6%95%B8%E4%B8%AD%E4%BD%BF%E7%94%A8%E5%B7%A2%E7%8B%80%20JSON%20%E7%B5%90%E6%A7%8B%E3%80%82

williamchong (approved):

PR #574, opened 2023-12-08
[TAS-863] ✨ `GET /book/store/list` supports pagination

WeiJun0827:


src/routes/likernft/book/store.ts
```diff
@@ -22,10 +22,20 @@ const router = Router();
...
   try {
-    const { wallet } = req.query;
+    const {
+      wallet,
+      before,
+      limit,
+      key,
+    } = req.query;
+    const pagination = {
+      before: before ? Number(before) : Date.now(),
```
williamchong: Maybe round `before` to nearest minute and add cache header? User don't need ms level precision update
Is `before` supposed to be ms or s?

src/routes/likernft/book/store.ts
```diff
@@ -22,10 +22,20 @@ const router = Router();
...
+      limit: limit ? Number(limit) : 10,
```
williamchong: Prefer bounding input params like `limit` in `/routes` instead of `listLatestNFTBookInfo` since we might use `listLatestNFTBookInfo` internally else where

src/util/api/likernft/book/index.ts
```diff
@@ -130,16 +130,45 @@ export async function getNftBookInfo(classId) {
...
+}: {
+  before?: number;
+  limit?: number;
+  key?: number | null;
+} = {}) {
+  if (limit > 100) throw new ValidationError('LIMIT_TOO_LARGE');
+  const timestamp = Timestamp.fromMillis(key || before);
+  const query = await likeNFTBookCollection
+    .orderBy('timestamp', 'desc')
+    .startAfter(timestamp)
```
williamchong: Instead of a dynamic default value of `Date.now()`, how about don't add `startAfter` if neither `before` or `key` is set

src/util/api/likernft/book/index.ts
```diff
@@ -130,16 +130,45 @@ export async function getNftBookInfo(classId) {
...
+  before?: number;
+  limit?: number;
+  key?: number | null;
+} = {}) {
+  if (limit > 100) throw new ValidationError('LIMIT_TOO_LARGE');
+  const timestamp = Timestamp.fromMillis(key || before);
+  const query = await likeNFTBookCollection
+    .where('ownerWallet', '==', ownerWallet)
+    .orderBy('timestamp', 'desc')
+    .startAfter(timestamp)
```
williamchong: Instead of a dynamic default value of `Date.now()`, how about don't add `startAfter` if neither `before` or `key` is set

src/routes/likernft/book/store.ts
WeiJun0827: `Timestamp` class only provides `fromMillis` and `fromDate` functions.
Expect accepting ms as it's same as the derived timestamp format of `Date` class, and most common used.

Tho it's not a required precision, but Firebase can handle it anyway?

williamchong: On second thought it is more about making sure caller don't use a new `before` for every second so that we can have bette cache via nginx and cdn, which seems we can't handle here..

src/routes/likernft/book/store.ts
```diff
@@ -45,14 +56,17 @@ router.get('/list', jwtOptionalAuth('read:nftbook'), async (req, res, next) => {
...
       };
       if (req.user && req.user.wallet === wallet) {
         result.pendingNFTCount = pendingNFTCount;
         result.sold = sold;
       }
       return result;
     });
-    res.json({ list });
+    const nextKey = list.length < conditions.limit ? null : list[list.length - 1].timestamp;
+    res.set('Cache-Control', 'public, max-age=6, s-maxage=6, stale-if-error=60');
```
williamchong:
```suggestion
    res.set('Cache-Control', 'public, max-age=60, s-maxage=60, stale-if-error=600');
```
This is not block related, we can have longer cache time

williamchong (approved):

src/routes/likernft/book/store.ts
williamchong: On second thought, we need to set `private` for `isAuthorized`

williamchong:
```suggestion
    if (isAuthorized) {
      setNoCacheHeader(res)
    } else {
      res.set('Cache-Control', 'public, max-age=60, s-maxage=60, stale-if-error=600');
    }
```
On second thought, we need to set `private` for `isAuthorized`, sth like this
Also this is not block time related, we can set much longer cache time

williamchong (changes requested):

src/routes/likernft/book/store.ts
```diff
@@ -35,24 +46,33 @@ router.get('/list', jwtOptionalAuth('read:nftbook'), async (req, res, next) => {
...
       }
       return result;
     });
-    res.json({ list });
+    const nextKey = list.length < conditions.limit ? null : list[list.length - 1].timestamp;
+    if (allAuthorized) {
+      res.set('Cache-Control', 'public, max-age=60, s-maxage=60, stale-if-error=600');
+    } else {
+      res.set('Cache-Control', 'no-store');
+    }
```
williamchong: the condition should be flip, we should only cache resposne without any auth
or we can simple check for `req.user` not exist?
Since `jwtOptionalAuth` set no cache already

```suggestion
    if (!req.user) {
      res.set('Cache-Control', 'public, max-age=60, s-maxage=60, stale-if-error=600');
    }
```

williamchong (changes requested):

src/routes/likernft/book/store.ts
WeiJun0827: OK, I See, should like this?

williamchong (approved):

PR #578, opened 2023-12-15
[TAS-823] ✨ Add book store collection API

williamchong:
Base on https://github.com/likecoin/likecoin-api-public/pull/445

Implement book store collection purchase logic


WeiJun0827 (approved):
Should we also provide any cache header?

PR #580, opened 2023-12-19
🔧 Set `moduleResolution` to `node16`

WeiJun0827:
Prerequisite config of using w3up client

### DIFF
--- a/src/util/arweave/index.ts
+++ b/src/util/arweave/index.ts
@@ -1,5 +1,4 @@
 import Arweave from 'arweave/node';
-import Bundlr from '@bundlr-network/client';
 import BigNumber from 'bignumber.js';
 import stringify from 'fast-json-stable-stringify';
 import LRU from 'lru-cache';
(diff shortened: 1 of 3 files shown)


WeiJun0827:
Demo video

https://github.com/likecoin/likecoin-api-public/assets/33746295/bc19f3d7-f41d-4041-9c1f-0e9e28430d47

src/util/arweave/index.ts
```diff
@@ -30,11 +29,22 @@ const arweaveGraphQL = Arweave.init({
...
-  'arweave',
-  jwk,
-);
+// eslint-disable-next-line no-underscore-dangle
+let _bundlr;
+
+async function getBundlr() {
+  if (!_bundlr) {
+    // eslint-disable-next-line global-require
+    global.crypto = require('crypto'); // hack for bundlr
```
williamchong:
```suggestion
    if (!global.crypto) global.crypto = require('crypto'); // hack for bundlr
```

src/util/arweave/signer.ts
```diff
@@ -1,14 +1,24 @@
...
-  'matic',
-  BUNDLR_MATIC_WALLET_PRIVATE_KEY,
-);
+// eslint-disable-next-line no-underscore-dangle
+let _maticBundlr;
+
+export async function getMaticBundlr() {
+  if (!_maticBundlr) {
+    // eslint-disable-next-line global-require
+    global.crypto = require('crypto'); // hack for bundlr
```
williamchong:
```suggestion
    if (!global.crypto) global.crypto = require('crypto'); // hack for bundlr
```

williamchong (changes requested):

williamchong (approved):

PR #582, opened 2023-12-21
🐛 Fix log price format error

WeiJun0827:


src/util/api/likernft/book/purchase.ts
```diff
@@ -493,6 +493,10 @@ export async function processNFTBookStripePurchase(
       isGift,
     });
 
+    let priceInCurrency = price;
+    if (defaultPaymentCurrency === 'HKD') {
+      priceInCurrency = Math.round((price * USD_TO_HKD_RATIO) / 10) * 10;
+    }
```
nwingt: Can we refactor [this](https://github.com/likecoin/likecoin-api-public/blob/master/src/util/api/likernft/book/purchase.ts#L268-L273) into function and apply it in here?

src/util/api/likernft/book/purchase.ts
```diff
@@ -493,6 +496,7 @@ export async function processNFTBookStripePurchase(
       isGift,
     });
 
+    const priceInCurrency = convertUSDToCurrency(price, defaultPaymentCurrency || 'USD');
```
nwingt:
```suggestion
    const convertedCurrency = defaultPaymentCurrency === 'HKD' ? 'HKD' : 'USD';
    const convertedPriceInDecimal = convertUSDToCurrency(price, convertedCurrency);
```

src/util/api/likernft/book/purchase.ts
```diff
@@ -513,8 +517,8 @@ export async function processNFTBookStripePurchase(
         paymentId,
         email,
         priceName,
-        priceWithCurrency: `${price} ${defaultPaymentCurrency || 'USD'}`,
-        method: 'USD',
+        priceWithCurrency: `${priceInCurrency} ${defaultPaymentCurrency || 'USD'}`,
```
nwingt:
```suggestion
        priceWithCurrency: `${convertedPriceInDecimal} ${convertedCurrency}`,
```

nwingt (changes requested):

nwingt (approved):

PR #589, opened 2024-01-03
[TAS-880] ✨ Support storing auto delivering NFTs in API wallet

WeiJun0827:
- Add `isAutoDeliver` boolean in each price(edition)
- User can set `isAutoDeliver` when creating an NFT Book or adding a new price
- `isAutoDeliver` is read-only after created
- Stock of auto delivering price can not be modified


williamchong (approved):
While this approach work, considering accepting a txHash in payload so that we can check user did sent a successful transaction on related autodeliver class and NFT to API wallet.
This might also allow us to skip the `classIdRef.collection('nft').get()` but it is not a huge issue for now

src/util/api/likernft/book/index.ts
```diff
@@ -127,7 +163,38 @@ export async function updateNftBookInfo(classId: string, {
   if (mustClaimToView !== undefined) { payload.mustClaimToView = mustClaimToView; }
   if (hideDownload !== undefined) { payload.hideDownload = hideDownload; }
   if (canPayByLIKE !== undefined) { payload.canPayByLIKE = canPayByLIKE; }
-  await likeNFTBookCollection.doc(classId).update(payload);
+  const classIdRef = likeNFTBookCollection.doc(classId);
+  let newNFTIds: string[] = [];
+  if (apiWalletOwnedNFTIds.length) {
+    const existingNFTIds = (await classIdRef.collection('nft').get()).docs.map((d) => d.id);
```
williamchong: Consider filtering by isSold: false and limit size to apiWalletOwnedNFTIds.length

src/routes/likernft/book/store.ts
```diff
@@ -223,7 +225,8 @@ router.post(['/:classId/price/:priceIndex', '/class/:classId/price/:priceIndex']
   try {
     const { classId, priceIndex: priceIndexString } = req.params;
     const priceIndex = Number(priceIndexString);
-    const { price } = req.body;
+    const { price, sendNFTsToAPIWalletTxHash } = req.body;
```
williamchong: something shorter like `autoDeliverNFTsTxHash`？

WeiJun0827: sure~

PR #610, opened 2024-01-17
🔥 Remove deprecated web3.storage

williamchong:
Removes a deprecated factor when debugging arweave upload issue


WeiJun0827:
post-approved

PR #636, opened 2024-03-07
[TAS-1158] 👔 Refactor stripe destination charge to separate transfer

williamchong:
@nwingt might need to change analytics script after this PR, if we managed to get transfer working without application fee

This PR also changed the behaviour of fee collecting for channels. Old behaviour is to send them to author, new behaviour this will be held in our stripe account also. This allow us to do payout in stripe connect later after channel register a stripe connect account.

Ref:
Before: https://docs.stripe.com/connect/destination-charges
After: https://docs.stripe.com/connect/separate-charges-and-transfers


nwingt (approved):
Please resolve conflict

PR #657, opened 2024-04-02
🚸 Sending book sales email by default

AuroraHuang22:


src/util/api/likernft/book/purchase.ts
```diff
@@ -690,7 +690,7 @@ export async function sendNFTBookPurchaseEmail({
       mustClaimToView,
     });
   }
-  if (notificationEmails && notificationEmails.length) {
+  if (notificationEmails) {
```
williamchong: if we don't have `notificationEmails` we should pass `notificationEmails = []`
Also do we know if aws ses accept 
```Destination: {
      ToAddresses: [],
      BccAddresses...
    },
```
as valie input?

williamchong (changes requested):

src/util/api/likernft/book/purchase.ts
AuroraHuang22: Since the Toaddresses is [not required](https://docs.aws.amazon.com/ses/latest/APIReference/API_Destination.html) for ses:

[🎨 Handle empty toAddress for ses](https://github.com/likecoin/likecoin-api-public/pull/657/commits/bcdc39caed215cd72bb8fc2f7fe11ffe48eba5f6)

williamchong (approved):

PR #660, opened 2024-04-07
[TAS-1518] 🧑‍💻 Return timestamp in NFT collection

nwingt:


src/util/api/likernft/collection/index.ts
```diff
@@ -172,6 +174,7 @@ export async function createNFTCollectionByType(
     image,
     type,
     typePayload,
+    timestamp: createdDoc.data().timestamp,
```
williamchong: Prefer to name the `.data()` as variable first since its a function call

nwingt: Updated

williamchong (approved):

PR #662, opened 2024-04-09
[TAS-1523] ✨Add Airtable record for each new NFT Book/Collection

nwingt:


src/util/airtable.ts
```diff
@@ -0,0 +1,81 @@
...
+      fields['Owner Name'] = ownerData.displayName;
+    }
+
+    await base('Publications').create([{ fields }]);
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err);
+  }
+
+  return Promise.resolve();
```
williamchong: We don't need to return anything if not needed

williamchong (changes requested):

src/util/airtable.ts
```diff
@@ -0,0 +1,81 @@
...
+  id: string;
+  ownerWallet: string;
+  type: string;
+  minPrice: number;
+  maxPrice: number;
+  imageURL: string;
+  iscnIdPrefix?: string;
+  iscnVersionAtMint?: number;
+}): Promise<void> {
+  if (TEST_MODE) return Promise.resolve();
```
williamchong:
```suggestion
  if (TEST_MODE || ! base) return;
```

williamchong (changes requested):

src/util/airtable.ts
nwingt: How about checking `!base` only?

williamchong (approved):

src/util/airtable.ts
williamchong: Sure

williamchong (approved):

PR #664, opened 2024-04-10
[TAS-1455] ✨ Add search API for bookstore

williamchong:


src/util/airtable.ts
```diff
@@ -143,7 +143,8 @@ export async function queryAirtableForPublication({ query }) {
         type,
         id,
         name,
-        imageUrl,
+        url: `https://${LIKER_LAND_HOSTNAME}/nft/${(id as string).startsWith('col_book') ? 'collection' : 'class'}/${id}`,
```
nwingt: I think `url` should be a formula field in Airtable

src/util/airtable.ts
```diff
@@ -143,7 +143,8 @@ export async function queryAirtableForPublication({ query }) {
...
+        imageUrl: (imageUrl as string)?.replace('ar://', 'https://arweave.net/'),
```
nwingt: `imageUrl` should be converted to `https://` when creating a record

You can also use [`parseImageURLFromMetadata()`](https://github.com/likecoin/likecoin-api-public/blob/3b8b45e1a0f341585fa4221d3641a0c1bdc1a364/src/util/api/likernft/metadata.ts#L49)

src/util/airtable.ts
```diff
@@ -94,4 +94,68 @@ export async function createAirtablePublicationRecord({
...
+        maxPrice,
+        description,
+        iscnId,
+      }));
+    return result;
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err);
+  }
+  return null;
```
nwingt, line 158: Should we return an empty array?

williamchong: null to tell it is an error case instead of real empty result

nwingt (approved):

PR #669, opened 2024-04-16
[TAS-1522] ✨ Add Airtable record for each sales

nwingt:


src/util/airtable.ts
```diff
@@ -158,4 +169,199 @@ export async function queryAirtableForPublication({ query }) {
...
+  let paymentMethod = 'unknown';
+  let balanceTxAmount = 0;
+  let balanceTxCurrency: string | undefined;
+  let balanceTxExchangeRate: number | undefined;
+  let balanceTxNetAmount = 0;
+  let stripeFee = 0;
+  let stripeFeeCurrency = 'usd';
+  let feeTotal = 0;
+
+  if (pi.latest_charge && typeof pi.latest_charge !== 'string') {
```
williamchong: Should we fetch object from stripe if this is string?

src/util/airtable.ts
```diff
@@ -158,4 +169,199 @@ export async function queryAirtableForPublication({ query }) {
...
+    const paymentMethodDetails = pi.latest_charge?.payment_method_details;
+    if (paymentMethodDetails) {
+      paymentMethod = paymentMethodDetails.type;
+      if (paymentMethod === 'card' && paymentMethodDetails.card?.brand) {
+        paymentMethod = paymentMethodDetails.card.brand;
+      }
+    }
+
+    const balanceTx = pi.latest_charge.balance_transaction;
+    if (balanceTx && typeof balanceTx !== 'string') {
```
williamchong: Should we fetch object from stripe if this is string?

src/util/airtable.ts
```diff
@@ -158,4 +169,199 @@ export async function queryAirtableForPublication({ query }) {
...
+
+      balanceTx.fee_details.forEach(({
+        type,
+        amount,
+        currency,
+      }) => {
+        switch (type) {
+          case 'stripe_fee':
+            stripeFee = amount / 100;
+            stripeFeeCurrency = currency;
```
williamchong: Should use .find instead?

williamchong (changes requested):

src/util/airtable.ts
nwingt: It shouldn't be a string as we added `expand` param in [here](https://github.com/likecoin/likecoin-api-public/blob/6e769f6a83ddf71a86df1fcc6410bf2330bcbcba/src/util/api/likernft/book/collection/purchase.ts#L461-L462)

src/util/airtable.ts
nwingt: Updated

src/util/airtable.ts
williamchong: Add a console error if we encounter string? Someone might misuse this function without knowing

williamchong (approved):

PR #675, opened 2024-04-23
[TAS-1449] ✨ Add transfer_group support

williamchong:


src/util/airtable.ts
```diff
@@ -293,23 +307,43 @@ function normalizeStripePaymentIntentForAirtableBookSalesRecord(pi: Stripe.Payme
...
+  } else {
+    likerLandFee = balanceTxAmount * NFT_BOOK_LIKER_LAND_FEE_RATIO;
+  }
 
-  const payableAmount = !withStripeConnect
-    ? balanceTxAmount - stripeFee - likerLandFee - likerLandCommission
-      - likerLandArtFee - likerLandTipFee
-    : 0;
+  // NOTE: We have to collect commission for tx with Stripe Connect before the commission fix date
+  const receivableAmount = isLikerLandChannel
```
nwingt: Since this field is deprecated, can we just set it to 0? it returns boolean now

nwingt (changes requested):

src/util/airtable.ts
williamchong: Seems 2 lines when missing after rebase, fixed

PR #679, opened 2024-05-06
[TAS-1618] 🐛 Calculate all fees with respect to the exchange rate.

nwingt:


src/util/airtable.ts
```diff
@@ -268,10 +286,8 @@ function normalizeStripePaymentIntentForAirtableBookSalesRecord(pi: Stripe.Payme
 
   const otherCommission = !withStripeConnect && !isLikerLandChannel ? channelCommission : 0;
 
-  // NOTE: Liker Land commission is included in the application fee after the commission fix date
-  const applicationFee = pi.application_fee_amount ? pi.application_fee_amount / 100 : 0;
   const likerLandFee = withStripeConnect
-    ? applicationFee - stripeFee - likerLandCommission - likerLandArtFee - likerLandTipFee
+    ? applicationFeeAmount - stripeFee - likerLandCommission - likerLandArtFee - likerLandTipFee
     : balanceTxAmount * NFT_BOOK_LIKER_LAND_FEE_RATIO;
```
williamchong, line 300: I think not only `applicationFee`, but `likerLandCommission`, `likerLandArtFee`, `likerLandTipFee`, all data that come from `pi` or `metadata` without explicit currency stated should all be converted

williamchong (changes requested):

src/util/airtable.ts
nwingt: Seem Stripe fee currency is always USD, but I guard it anyway

src/util/api/likernft/book/collection/purchase.ts
```diff
@@ -458,7 +458,7 @@ export async function processNFTBookCollectionStripePurchase(
       originalPriceInDecimal,
     } = txData;
     const [capturedPaymentIntent, collectionData] = await Promise.all([
-      stripe.paymentIntents.capture(paymentIntent as string, { expand: ['latest_charge.balance_transaction'] }),
+      stripe.paymentIntents.capture(paymentIntent as string, { expand: ['latest_charge.balance_transaction', 'latest_charge.application_fee'] }),
```
williamchong: You might also want to edit the one in `src/util/api/likernft/book/purchase.ts`

williamchong (changes requested):

src/util/api/likernft/book/collection/purchase.ts
nwingt: Updated

williamchong (approved):
`balanceTxExchangeRate` could be `1` instead of `undefined`? Not sure which default value is best

nwingt:
If the default is set to 1, then it still multiplies by 1, which I think is not necessary

PR #687, opened 2024-06-10
✨ Add hideDownload & excluded wallet for bookstore list 

nwingt:


src/routes/likernft/book/store.ts
```diff
@@ -43,12 +43,14 @@ router.get('/list', jwtOptionalAuth('read:nftbook'), async (req, res, next) => {
   try {
     const {
       wallet,
+      excluded_wallet: excludedWallet,
```
williamchong:
```suggestion
      exclude_wallet: excludedWallet,
```
Using verb seems better for qs

src/util/api/likernft/book/index.ts
```diff
@@ -230,6 +231,7 @@ export async function listLatestNFTBookInfo({
 } = {}) {
   let snapshot = likeNFTBookCollection.orderBy('timestamp', 'desc');
   if (ownerWallet) snapshot = snapshot.where('ownerWallet', '==', ownerWallet);
+  if (excludedOwnerWallet) snapshot = snapshot.where('ownerWallet', '!=', excludedOwnerWallet);
```
williamchong, line 235: Might need to build index

src/routes/likernft/book/store.ts
nwingt: But are `wallet`, `limit`, and `key` also a noun?

src/util/api/likernft/book/index.ts
nwingt: built

src/routes/likernft/book/store.ts
williamchong: it's like `ignore_whitespace=1` vs `ignored_whitespace=1`

williamchong (approved):

PR #688, opened 2024-06-13
🐛 Should use pi currency instead of balance tx currency in connect logic

williamchong:
only new data would be fix, old data need to manually fix 1 by 1


src/util/api/likernft/book/purchase.ts
```diff
@@ -993,7 +993,7 @@ export async function processNFTBookStripePurchase(
         amountTotal,
         chargeId,
         exchangeRate,
-        currency,
+        currency: defaultPaymentCurrency,
```
nwingt: Why set `defaultPaymentCurrency` here?

williamchong: actually we should not ..

nwingt (approved):

PR #698, opened 2024-06-27
✨ Add DRM-free field in Publication for Airtable

nwingt:


williamchong (approved):
Now sure drm free should be default true or false

nwingt:
Follow this https://github.com/likecoin/likecoin-api-public/blob/99072eb02345078e66af15d1fe342e53f57fd468/src/routes/likernft/book/store.ts#L474

PR #700, opened 2024-07-05
[TAS-1753] ✨ Add cart API for books

williamchong:
Requires https://github.com/likecoin/likecoin-api-public/pull/699


src/util/api/likernft/book/cart.ts
```diff
@@ -0,0 +1,800 @@
...
+      collectionId,
+      coupon,
+      customPriceInDecimal,
+      quantity = 1,
+      from: itemFrom,
+    } = item;
+    let info;
+    if (!Number.isInteger(quantity) || quantity <= 1) {
+      throw new ValidationError('QUANTITY_INVALID');
+    }
```
AuroraHuang22:
```suggestion
    if (!Number.isInteger(quantity) || quantity < 1) {
      throw new ValidationError('QUANTITY_INVALID');
    }
```
If the quantity =1 , it will also throw an error

AuroraHuang22 (changes requested):

AuroraHuang22 (approved):

PR #701, opened 2024-07-05
[TAS-1835] ✨ Add book cart claim handler

williamchong:
Requires https://github.com/likecoin/likecoin-api-public/pull/700


src/util/api/likernft/book/cart.ts
```diff
@@ -0,0 +1,851 @@
...
+          from,
+        }),
+        createAirtableBookSalesRecordFromStripePaymentIntent({
+          pi: capturedPaymentIntent,
+          quantity,
+          feeInfo,
+          transfers,
+          shippingCountry: undefined,
+          shippingCost: undefined,
+        }),
```
nwingt: Since Stripe pi metadata doesn't have `classId` or `collectionId` for cart, we should explicitly pass them here

nwingt (approved):

PR #704, opened 2024-07-10
🐛 Fix incorrect payable amount for Airtable

nwingt:


src/util/airtable.ts
```diff
@@ -350,16 +345,25 @@ function normalizeStripePaymentIntentForAirtableBookSalesRecord(
     likerLandFee = balanceTxAmount * NFT_BOOK_LIKER_LAND_FEE_RATIO;
   }
 
-  const hasPaidChannelCommission = isLikerLandChannel || !!transfers?.find((t) => t.metadata?.type === 'channelCommission');
```
williamchong, line 353: We probably can keep this

williamchong (approved):

PR #708, opened 2024-07-18
💬 Change commission email title and add paymentId

williamchong:


nwingt (changes requested):
I think either using `US$` or `USD`

nwingt (approved):

AuroraHuang22 (approved):

PR #716, opened 2024-08-06
📈 Add referrer for GET book purchase

williamchong:


nwingt (approved):

nwingt:
Should we log referrer to Airtable too?

PR #717, opened 2024-08-06
[TAS-959] ✨ Support auto deliver in book collection purchase

williamchong:


src/util/api/likernft/collection/index.ts
```diff
@@ -229,7 +341,64 @@ export async function patchNFTCollectionById(
...
+          {
+            nftId,
+            classId,
+            isSold: false,
+            isProcessing: false,
+            timestamp: FieldValue.serverTimestamp(),
+          },
+          { merge: true },
+        );
+      }
```
nwingt, line 397: Seems can be refactored

nwingt (approved):

PR #734, opened 2024-08-26
👔 Set default payout interval for new connected accounts

williamchong:


src/routes/likernft/book/user.ts
```diff
@@ -147,6 +147,15 @@ router.post(
           metadata: {
             wallet,
           },
+          settings: {
+            payouts: {
+              schedule: {
+                interval: 'monthly',
+                monthly_anchor: 8,
+                delay_days: 14,
```
nwingt: Should be 7?

williamchong: It depends if we expect any refund and charge back

nwingt: then change monthly_anchor to 15?

williamchong: lets do 7 for now then

nwingt (approved):

PR #738, opened 2024-08-30
[TAS-2123] ✨ Add book user payout apis

williamchong:


src/routes/likernft/book/user.ts
```diff
@@ -223,6 +223,110 @@ router.post(
...
+      }
+      const {
+        stripeConnectAccountId,
+        isStripeConnectReady,
+      } = userData;
+      if (!stripeConnectAccountId) {
+        throw new ValidationError('ACCOUNT_NOT_CREATED', 404);
+      }
+      if (!isStripeConnectReady) {
+        throw new ValidationError('USER_NOT_COMPLETED_ONBOARD', 405);
```
nwingt: Should be 406?

williamchong: 406 is used for `Accept` and `Accept-Encoding` related error, maybe just 400 or 409

nwingt (approved):

PR #742, opened 2024-09-03
[TAS-2146] 🚸 Fetch likerland & authcore info in user command

williamchong:
Require https://github.com/likecoin/liker-land/pull/1775 for new likerland api

### DIFF
--- a/src/routes/slack/user.ts
+++ b/src/routes/slack/user.ts
@@ -1,0 +1,96 @@
+import { Router } from 'express';
+import { slackTokenChecker } from '../../middleware/slack';
+import {
+  SLACK_COMMAND_TOKEN,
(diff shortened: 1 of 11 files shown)


nwingt (approved):

src/routes/slack/user.ts
```diff
@@ -0,0 +1,145 @@
...
+      } catch (err) {
+        // eslint-disable-next-line no-console
+        console.error(err);
+      }
+    }
+    Object.assign(userInfo, userData, civicInfo);
+  }
+
+  const attachments: any = [{
+    text: Object.keys(userInfo).map((key) => `${key}: ${userInfo[key]}`).join('\n'),
```
nwingt, line 91: Can remove text if we have attachments?
image

PR #760, opened 2024-09-19
[TAS-2263] 📈 Add gad_source support

williamchong:


nwingt (approved):
Started to think we should encapsulate these params into a single object, e.g., `whitelistedQuery`

williamchong:
> Started to think we should encapsulate these params into a single object, e.g., `whitelistedQuery`

Maybe just pass anything that exists in qs for `POST` requests

Not sure if any security risk, hard to imagine

PR #764, opened 2024-09-24
[TAS-2289] 📈 Add crisp user segment from server purchase events

williamchong:


src/util/crisp.ts
```diff
@@ -0,0 +1,70 @@
...
+        nickname: displayName || wallet || email.split('@')[0],
+      },
+      segments,
+      active: Math.floor(Date.now() / 1000),
+    });
+  }
+  await CrispClient.website.updatePeopleData(CRISP_WEBSITE_ID, email, {
+    like_wallet: wallet,
+    login_method: loginMethod,
+  });
```
nwingt, line 69: Why do we call this separately?
Can this be merged with the `addPeopleData`/`updatePeopleData` above?

src/util/api/likernft/book/purchase.ts
```diff
@@ -1378,6 +1380,20 @@ export async function processNFTBookStripePurchase(
         stripeFeeAmount,
       }),
     ]);
+
+    if (email) {
+      const segments = ['purchaser'];
+      if (feeInfo.customPriceDiff) segments.push('tipper');
+      if (CRISP_WALLET_TO_SEGMENT_MAPPING?.[ownerWallet]) {
+        segments.push(CRISP_WALLET_TO_SEGMENT_MAPPING[ownerWallet]);
+      }
```
nwingt: it seems that we can refactor this part

williamchong: You mean wrap these three lines as a function in util/crisp?

nwingt: Yes

nwingt (approved):

PR #790, opened 2024-10-10
[TAS-2399] ✨ Add out of stock notification

williamchong:


nwingt:
Not sure if we want to have another alert when stock < 5

williamchong:
> Not sure if we want to have another alert when stock < 5

exactly 5 or every time lower than 5?

williamchong:
> Not sure if we want to have another alert when stock < 5

updated

nwingt (approved):

PR #798, opened 2024-10-21
[TAS-2422] 🎨 Include owner name in search fields

AuroraHuang22:


src/util/airtable.ts
```diff
@@ -135,6 +135,7 @@ export async function queryAirtableForPublication({ query }) {
     const formulas = [
       'Name',
       'Description',
+      'Owner Name',
```
nwingt, line 138: Maybe add `Author` too?

AuroraHuang22: [🎨 Include 'Author' field in search](https://github.com/likecoin/likecoin-api-public/pull/798/commits/5296b8222b3138610cce059be15c560c30ecd791)

williamchong (approved):

PR #800, opened 2024-10-22
[TAS-1968] ✨ Add Arweave link APIs

williamchong:


src/routes/arweave/index.ts
```diff
@@ -130,6 +168,42 @@ router.post(
...
+      const {
+        arweaveId, token: docToken, isRequireAuth, ownerWallet,
+      } = tx;
+      if (isRequireAuth
+        && (
+          req.user?.wallet !== ownerWallet
+          && token !== docToken
+          && (ARWEAVE_LINK_INTERNAL_TOKEN && token !== ARWEAVE_LINK_INTERNAL_TOKEN))
+      ) {
+        throw new ValidationError('MISSING_USER', 401);
```
nwingt: Should we throw different errors for each condition?

src/routes/arweave/index.ts
```diff
@@ -103,12 +117,36 @@ router.post(
...
+      if (!arweaveId) throw new ValidationError('MISSING_ARWEAVE_ID');
+      if (isRequireAuth && !req.user?.wallet) throw new ValidationError('MISSING_USER', 401);
+      const tx = await getArweaveTxInfo(txHash);
+      if (!tx) throw new ValidationError('TX_NOT_FOUND', 404);
+      const { ownerWallet, authToken } = tx;
+      const userWallet = req.user?.wallet || '';
+      const isAuthed = (ownerWallet && userWallet === ownerWallet)
+        || (authToken && authToken === token);
+      if (!isAuthed) throw new ValidationError('INVALID_TOKEN', 403);
+      if (tx.status !== 'pending') throw new ValidationError('TX_ALREADY_REGISTERED', 403);
```
nwingt:
```suggestion
      if (tx.status !== 'pending') throw new ValidationError('TX_ALREADY_REGISTERED', 409);
```

src/routes/arweave/index.ts
williamchong: Updated, but the condition is valid token or valid user, seems hard to return if its a 403 for token or user in a simple way

nwingt (approved):

PR #809, opened 2024-11-14
[TAS-2591] ✨ Add Google Retail product import

williamchong:


src/util/googleRetail.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export async function importProductFromBookListing(bookData) {
+  const products = formatBookData(bookData);
+  await importProductToRetailCatalog(products);
+}
+
+export async function importProductFromCollection(collectionData) {
+  const products = formatCollectionData(collectionData);
+  await importProductToRetailCatalog(products);
+}
```
nwingt:
```suggestion
export async function importGoogleRetailProductFromBookListing(bookData) {
  const products = formatBookData(bookData);
  await importProductToRetailCatalog(products);
}

export async function importGoogleRetailProductFromCollection(collectionData) {
  const products = formatCollectionData(collectionData);
  await importProductToRetailCatalog(products);
}
```

nwingt (approved):

PR #811, opened 2024-11-16
[TAS-2407] 👔 Add reply to CS for sales email

nwingt:


williamchong (approved):
cs@liker.land is linked to crisp?

nwingt:
> [cs@liker.land](mailto:cs@liker.land) is linked to crisp?

Yes

PR #814, opened 2024-11-25
[TAS-2611] ✨ Auto fetch evmWallet for authcore users

williamchong:


williamchong:
considering support other oauth except authcore during migration

PR #840, opened 2025-01-13
🐛 Fix wrong sales notification tip amount for quantity > 1 sale

### LINKED ISSUE #263: buy multiple copies misinterpreted as tipping
Screenshot 2025-01-13 at 16 27 07

this order is for buying 5 copies (5 x 0.99 = 4.95) instead of tipping 3.96

williamchong:
Fixes https://github.com/likecoin/nft-book-press/issues/263

### DIFF
--- a/src/util/ses.ts
+++ b/src/util/ses.ts
@@ -976,6 +976,7 @@ export function sendNFTBookSalesEmail({
   buyerEmail,
   bookName,
   amount,
+  quantity,
   originalPrice,
   phone,
   shippingDetails,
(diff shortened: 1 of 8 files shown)


src/util/ses.ts
```diff
@@ -987,10 +988,11 @@ export function sendNFTBookSalesEmail({
   <br/>
   <p>Congratulation!</p>
   <p>${buyerEmail} has bought your book ${bookName} for $${amount}${isGift ? ` as a gift to ${giftToName}<${giftToEmail}>` : ''}.</p>`;
```
nwingt, line 990: should we also show quantity here?

williamchong: maybe we should just always show the `Price:` row

nwingt (approved):

AuroraHuang22 (approved):

PR #879, opened 2025-03-10
💩 Claim cart items in seq to avoid race tx seq

williamchong:


nwingt (approved):

nwingt:
Should we use transaction?

williamchong:
> Should we use transaction?

This is more about cosmos transaction was sent out of order and we don't have a dedicated queue/worker that manage cosmos txs

PR #931, opened 2025-05-29
👽 Handle v3 URL

nwingt:


src/util/liker-land.ts
```diff
@@ -106,6 +130,9 @@ export const getLikerLandNFTClassPageURL = ({
     qsPayload.gad_source = gadSource;
   }
   const qs = Object.entries(qsPayload).map(([key, value]) => `${key}=${value}`).join('&');
+  if (isV3) {
+    return getBook3URL(`/store/${classId}?${qs}`, { language });
+  }
   return getLikerLandURL(`/nft/class/${classId}?${qs}`, { language });
```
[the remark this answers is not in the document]
nwingt: In some scenarios, both the endpoint and the path changed

src/util/liker-land.ts
```diff
@@ -249,6 +278,9 @@ export const getLikerLandNFTClaimPageURL = ({
     qsPayload.gad_source = gadSource;
   }
   const qs = Object.entries(qsPayload).map(([key, value]) => `${key}=${value}`).join('&');
+  if (isV3) {
+    return getBook3URL(`/store/claim?${qs}`, { language });
+  }
   return getLikerLandURL(`/nft/claim?${qs}`, { language });
 };
```
williamchong: Maybe instead of boolean, we can use a string/enum that s has values like `rinkeby` `3ook` `sepolia`, then default value returns `liker.land`?

src/util/liker-land.ts
```diff
@@ -249,7 +291,13 @@ export const getLikerLandNFTClaimPageURL = ({
     qsPayload.gad_source = gadSource;
   }
   const qs = Object.entries(qsPayload).map(([key, value]) => `${key}=${value}`).join('&');
-  return getLikerLandURL(`/nft/claim?${qs}`, { language });
+  switch (site) {
+    case '3ook-com':
+      return getBook3URL(`/store/claim?${qs}`, { language });
+    case 'liker.land':
```
williamchong: Why `liker.land` is `.` but `3ook-com` is `-`?

williamchong (approved):

PR #939, opened 2025-06-06
[TAS-3878] 🐛 Fix book collection owner migration is missing

williamchong:


src/util/api/wallet/index.ts
```diff
@@ -118,6 +118,24 @@ async function migrateBookOwner(likeWallet: string, evmWallet: string) {
         }
       });
     });
+    await db.runTransaction(async (t) => {
+      const collectionQuery = await t.get(likeNFTCollectionCollection.where('ownerWallet', '==', likeWallet));
```
nwingt, line 122: should add evm chain filter?

PR #943, opened 2025-06-07
🚸 Return 429 for used arweave tx hash

williamchong:


src/routes/arweave/index.ts
```diff
@@ -95,11 +95,22 @@ router.post(
...
+      try {
+        ({ token } = await createNewArweaveTx(txHash, {
+          ipfsHash,
+          fileSize,
+          ownerWallet: req.user?.wallet || '',
+        }));
+      } catch (error) {
+        if ((error as Error)?.message.includes('ALREADY_EXISTS')) {
+          // eslint-disable-next-line no-console
+          console.error(error);
```
nwingt: should we use `console.info` instead of error as we have handled this?

williamchong: console.warn maybe, this is still not expected

PR #948, opened 2025-06-09
🔧 Remove commission query limit

williamchong:


src/util/api/wallet/index.ts
```diff
@@ -187,7 +187,8 @@ export async function migrateBookClassId(likeClassId: string, evmClassId: string
         t.get(likeNFTBookCollection.doc(likeClassId)),
         t.get(likeNFTBookCollection.doc(likeClassId)
           .collection('transactions')
-          .where('status', '!=', 'new')),
+          .where('status', 'in', ['paid', 'pendingNFT', 'completed'])
+          .limit(450)),
```
nwingt: What is this magic number
maybe add a note?

williamchong: Double checked turns out there is no 500 limit now
https://cloud.google.com/firestore/docs/release-notes#March_29_2023
Will update PR to not limit instead

PR #957, opened 2025-06-12
✨ Update for 3ook.com account registration

nwingt:


src/util/api/users/index.ts
```diff
@@ -285,6 +285,7 @@ async function userInfoQuery({
   email,
   platform,
   authCoreUserId,
+  magicUserId,
 }: {
   user: string;
```
williamchong:
```suggestion
  user?: string;
```

src/util/api/users/index.ts
```diff
@@ -370,6 +388,7 @@ export async function checkUserInfoUniqueness({
...
```
williamchong:
```suggestion
  user?: string;
```

williamchong (changes requested):

williamchong (approved):

PR #958, opened 2025-06-12
🚸 Show masked wallets for EMAIL_ALREADY_USED error

nwingt:


src/routes/users/registerLogin.ts
```diff
@@ -119,15 +119,13 @@ router.post(
...
-              payload.isEmailVerified = true;
+          if (magicDIDToken && magicUserId) {
+            const magicUserMetadata = await getMagicUserMetadataByDIDToken(magicDIDToken);
+            if (magicUserMetadata.issuer !== magicUserId) {
+              throw new ValidationError('MAGIC_USER_ID_MISMATCH');
             }
             payload.magicUserId = magicUserId;
+            payload.isEmailVerified = verifyEmailByMagicUserMetadata(email, magicUserMetadata);
           }
           payload.email = email;
```
williamchong: I think no need to post `magicUserId` in payload, checking email via `verifyEmailByMagicUserMetadata` should be enough to check if there is any inconsistency in client data

src/util/api/users/index.ts
```diff
@@ -193,12 +194,17 @@ export function checkEVMSignPayload({
...
     snapshot.forEach((doc) => {
       const docUser = doc.id;
       if (user !== docUser) {
-        throw new ValidationError('EMAIL_ALREADY_USED');
+        const { evmWallet, likeWallet } = doc.data();
+        const payload = isEmailVerified ? {
+          evmWallet: maskString(evmWallet),
+          likeWallet: maskString(likeWallet, { start: 11 }),
+        } : null;
+        throw new ValidationError('EMAIL_ALREADY_USED', 400, payload);
```
williamchong: Maybe using an `if` block can show more explicitly about the role `isEmailVerified` is playing here
```suggestion
        let payload = null;
        if (isEmailVerified) {
          const { evmWallet, likeWallet } = doc.data();
          payload = {
            evmWallet: maskString(evmWallet),
            likeWallet: maskString(likeWallet, { start: 11 }),
          };
        }
        throw new ValidationError('EMAIL_ALREADY_USED', 400, payload);
```

src/util/api/users/index.ts
```diff
@@ -386,6 +394,7 @@ export async function checkUserInfoUniqueness({
   likeWallet,
   evmWallet,
   email,
+  isEmailVerified = false,
```
williamchong: This should be in a third optional param, since in this existing named 2nd param, all the fields are acutally being checked for uniqueness (well maybe except `platform`, but it is still part of a key)
But `isEmailVerified` is just a modifier on the query result

Maybe sth like `checkUserInfoUniqueness(..., { isEmailVerified = false } = {})`

src/util/magic.ts
```diff
@@ -11,7 +12,28 @@ export async function getMagic(): Promise<Magic> {
...
+
+export function verifyEmailByMagicUserMetadata(
+  email: string,
+  magicUserMetadata: MagicUserMetadata,
+): boolean {
+  if (!magicUserMetadata.email) {
+    return false;
+  }
+  if (email !== magicUserMetadata.email) {
+    throw new ValidationError('MAGIC_EMAIL_MISMATCH');
```
williamchong: Not sure if it is a good idea to throw instead of just return false here, seems one would expect this to return false judging by function name

src/routes/users/registerLogin.ts
```diff
@@ -119,15 +119,13 @@ router.post(
...
-            if (magicUserMetadata.email) {
-              if (email !== magicUserMetadata.email) {
-                throw new ValidationError('MAGIC_EMAIL_MISMATCH');
-              }
-              payload.isEmailVerified = true;
+          if (magicDIDToken && magicUserId) {
+            const magicUserMetadata = await getMagicUserMetadataByDIDToken(magicDIDToken);
+            if (magicUserMetadata.issuer !== magicUserId) {
+              throw new ValidationError('MAGIC_USER_ID_MISMATCH');
             }
```
williamchong, line 126: Maybe also check if we need to throw for `magicUserMetadata.email !== email` here, and don't throw in `verifyEmailByMagicUserMetadata` for easier readability

src/routes/users/registerLogin.ts
nwingt: you mean not checking `magicUserId` or not storing it in user doc?

williamchong: not submitting `magicUserId` in body, not checking, just use `magicUserMetadata.issuer` from verified DID token to get `magicUserId` and store to use doc

src/util/magic.ts
```diff
@@ -21,13 +20,7 @@ export function verifyEmailByMagicUserMetadata(
...
   magicUserMetadata: MagicUserMetadata,
 ): boolean {
-  if (!magicUserMetadata.email) {
-    return false;
-  }
-  if (email !== magicUserMetadata.email) {
-    throw new ValidationError('MAGIC_EMAIL_MISMATCH');
-  }
-  return true;
+  return email === magicUserMetadata.email;
```
williamchong: Do we need `email && email === magicUserMetadata.email`, or is both undefined acceptable too?

nwingt: should be `!!magicUserMetadata.email && email === magicUserMetadata.email`

src/util/misc.ts
```diff
@@ -9,3 +9,14 @@ export function removeUndefinedObjectKey(obj) {
...
+
+export function maskString(
+  str: string | undefined,
+  {
+    start = 8,
+    end = 4,
+  } = {},
+): string | undefined {
+  if (!str || str.length <= start + end) return str;
+  return `${str.slice(0, start)}*****${str.slice(-end)}`;
```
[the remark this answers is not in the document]
nwingt: no need for dynamic

src/util/magic.ts
```diff
@@ -11,7 +11,22 @@ export async function getMagic(): Promise<Magic> {
...
   const magic = await getMagic();
-  return magic.users.getMetadataByIssuer(userId);
+  return magic.users.getMetadataByToken(didToken);
+}
+
+export function verifyEmailByMagicUserMetadata(
+  email: string,
+  magicUserMetadata: MagicUserMetadata,
+): boolean {
+  return !!magicUserMetadata.email && email === magicUserMetadata.email;
```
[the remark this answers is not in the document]
nwingt: What if email is undefined

williamchong (approved):
Why need to check for email first?

nwingt:
> Why need to check for email first?

I want it to return `EMAIL_ALREADY_USED` first, even `EVM_WALLET_ALREADY_EXIST`

PR #998, opened 2025-07-24
[TAS-4214] ✨ Add migrate EVM user via Magic email API

williamchong:


src/util/api/wallet/index.ts
```diff
@@ -15,6 +15,17 @@ import { createStripeProductFromNFTBookPrice } from '../likernft/book';
 import stripe from '../../stripe';
 import { bookCacheBucket } from '../../gcloudStorage';
 
+export async function findLikerByEmail(email: string): Promise<{
```
nwingt: maybe should put this function in `/util/api/users/index.ts` instead?

nwingt (approved):

PR #1003, opened 2025-07-30
[TAS-4191] 🔥 Remove crisp

williamchong:


src/util/ValidationHelper.ts
```diff
@@ -45,8 +45,6 @@ export function filterUserData(u) {
     isPhoneVerified,
     isEmailEnabled,
     authCoreUserId,
-    intercomToken,
```
nwingt, line 48: Why removing `intercomToken` too?

williamchong: it is never populated, will re-implement in the future in another way

PR #1005, opened 2025-07-31
[TAS-4306] ✨ Update airtable publication records on edit

williamchong:
Not sure if this is correct way to update airtable record

### DIFF
--- a/src/util/airtable.ts
+++ b/src/util/airtable.ts
@@ -11,6 +11,44 @@ import { getUserWithCivicLikerPropertiesByWallet } from './api/users';
 import { parseImageURLFromMetadata } from './api/likernft/metadata';
 import { TransactionFeeInfo } from './api/likernft/book/type';
 
+interface AirtablePublicationRecordParams {
+  name?: string | { zh: string, en: string };
+  description?: string | { zh: string, en: string };
+  iscnIdPrefix?: string;
+  ownerWallet?: string;
+  type?: string;
+  minPrice?: number;
+  maxPrice?: number;
+  imageURL?: string;
+  author?: string;
+  publisher?: string;
+  language?: string;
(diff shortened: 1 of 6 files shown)


src/util/airtable.ts
```diff
@@ -234,6 +234,130 @@ async function queryAirtablePublicationRecordById(id: string) {
...
+  keywords?: string[];
+  usageInfo?: string;
+  isbn?: string;
+  iscnIdPrefix?: string;
+  iscnObject?: any;
+  iscnContentMetadata?: any;
+  metadata?: any;
+  isDRMFree?: boolean;
+  isHidden?: boolean;
+}): Promise<void> {
```
nwingt: Can the refactor the type and share with `createAirtablePublicationRecord()`?

src/util/api/likernft/book/index.ts
```diff
@@ -365,6 +366,35 @@ export async function syncNFTBookInfoWithISCN(classId) {
     // eslint-disable-next-line no-console
     console.error(err);
   }
+
+  try {
+    const { ownerWallet } = bookInfo;
+    const minPrice = prices?.[0]?.priceInDecimal || 0;
+    const maxPrice = prices?.[prices.length - 1]?.priceInDecimal || 0;
```
nwingt: Please refer to this

https://github.com/likecoin/likecoin-api-public/blob/a16ffcdc7d0ba56ddb0f52904102daf393c06707/src/util/api/wallet/index.ts#L338-L339

src/util/airtable.ts
williamchong: Please check if you want to further merge the diff between optional and required parameters

nwingt (approved):

PR #1006, opened 2025-07-31
✨ Add user information to payload in wallet authorization

AuroraHuang22:


src/routes/wallet/index.ts
```diff
@@ -62,6 +62,14 @@ router.post('/authorize', async (req, res, next) => {
     } else {
       payload.likeWallet = inputWallet;
     }
+
+    const likerIdInfo = await getUserWithCivicLikerPropertiesByWallet(inputWallet);
+    if (likerIdInfo) {
+      payload.user = likerIdInfo.user;
+      if (!permissions.includes('read')) {
+        permissions.push('read');
```
williamchong: `read` means `read:*` which is a huge permission, should be explicitly requested by user instead of auto grant.
Also user should request explicit permission like `read:plus` instead of `read`

williamchong (changes requested):

src/routes/wallet/index.ts
williamchong: suggest to use `profile` and the `/profile` endpoint

williamchong (approved):

PR #1014, opened 2025-08-13
[TAS-4396] ✨ Add plus gift book API

williamchong:


src/util/api/plus/index.ts
```diff
@@ -40,28 +43,84 @@ export async function processStripeSubscriptionInvoice(
...
+        });
+      }
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Error creating gift cart from subscription:', error);
+    }
+  }
 
-  const customerId = subscription.customer as string;
+  const customerId = (customer as Stripe.Customer).id;
```
nwingt: Perhaps refactor customer?

nwingt (commented):
As we will have purchased gifts from one user to another

I think we could use, e.g. `freebie`, instead of `gift` for prefix to reduce ambiguity

williamchong:
> As we will have purchased gifts from one user to another
> 
> I think we could use, e.g. `freebie`, instead of `gift` for prefix to reduce ambiguity

Maybe product will later say plus gift can really be gifted to others..

PR #1043, opened 2025-09-05
[TAS-4523] 🔒️ Add intercom token on wallet auth

williamchong:


src/util/intercom.ts
```diff
@@ -0,0 +1,11 @@
+import jwt, { JwtPayload } from 'jsonwebtoken';
+import {
+  INTERCOM_API_SECRET,
+} from '../../config/config';
+
+export function createIntercomToken(payload: JwtPayload): string | undefined {
+  if (!INTERCOM_API_SECRET) return undefined;
+  return jwt.sign(payload, INTERCOM_API_SECRET, { expiresIn: '1h' });
```
nwingt, line 8: is 1h enough?

nwingt (approved):

src/util/intercom.ts
williamchong: seems this is use to exchange intercom cookie only, 1h is given in sample code

PR #1046, opened 2025-09-07
✨ Add intercom API

williamchong:


src/util/api/plus/index.ts
```diff
@@ -144,6 +145,19 @@ export async function processStripeSubscriptionInvoice(
...
+    await updateIntercomUserLikerPlusStatus({
+      userId: likerId,
+      isLikerPlus: true,
+      likerPlusPeriod: period,
+      likerPlusSince: since,
+      likerPlusCurrentPeriodEnd: currentPeriodEnd,
+    });
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error('Failed to update Intercom user LikerPlus status:', error);
```
nwingt: Since most of the code in `updateIntercomUserLikerPlusStatus()` is wrapped in a try-catch, I think this error message is unlikely to be logged.

src/util/api/users/register.ts
```diff
@@ -270,6 +271,20 @@ export async function handleUserRegistration({
...
+      userId: user,
+      email,
+      name: displayName,
+      signedUpAt: createObj.timestamp,
+      avatar: avatarURL,
+      evmWallet,
+    });
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error('Failed to create Intercom user:', error);
```
nwingt: Same here

PR #1051, opened 2025-09-09
✨ Update intercom user on liker ID migration

williamchong:


src/util/intercom.ts
```diff
@@ -68,6 +68,56 @@ export async function createIntercomUser({
   }
 }
 
+async function findIntercomContactByUserId(userId: string): Promise<string | null> {
```
nwingt:
```suggestion
async function findIntercomContactIdByUserId(userId: string): Promise<string | null> {
```

src/util/intercom.ts
```diff
@@ -68,6 +68,56 @@ export async function createIntercomUser({
...
+    },
+  });
+
+  const contact = searchResult.data?.[0];
+  return contact?.id || null;
+}
+
+async function updateIntercomContact(
+  contactId: string,
+  customAttributes: Record<string, unknown>,
```
nwingt: better use an interface? 
hard to know what are the fields

nwingt (approved):

PR #1054, opened 2025-09-11
✨ Support redirectClassId in listing

williamchong:


src/routes/likernft/book/store.ts
```diff
@@ -130,24 +131,31 @@ router.get('/list/moderated', jwtAuth('read:nftbook'), async (req, res, next) =>
...
-        ownerWallet,
-      };
-      return result;
-    });
+    const list = moderatedBookInfos
+      .filter((b) => {
+        const {
+          redirectClassId,
+        } = b;
+        return !redirectClassId;
```
AuroraHuang22: why not just
`.filter(b => b.redirectClassId == null)`

AuroraHuang22 (approved):

PR #1063, opened 2025-09-17
🏷️ Add indexer type from openapi spec

williamchong:


src/util/evm/nft.ts
```diff
@@ -11,6 +12,92 @@ import {
...
+interface PaginationResponse {
+  next_key: number;
+  count: number;
+}
+
+interface Account {
+  id: number;
+  cosmos_address?: string;
+  evm_address: string;
+  likeid?: string;
```
nwingt, line 24: is it typo?

williamchong: image
https://github.com/likecoin/likecoin-op/blob/main/likenft-indexer/openapi/schema.yaml#L754C9-L754C15
Somehow definition really includes this, but we could just drop it for now

nwingt (approved):

PR #1065, opened 2025-09-17
💥 Replace notificationEmails with ownerEmail

williamchong:


src/util/ses.ts
```diff
@@ -823,8 +823,8 @@ export function sendNFTBookOutOfStockEmail({
       },
     },
   };
-  if (emails && emails.length > 0) {
-    (params.Destination as any).ToAddresses = emails;
+  if (email) {
```
nwingt, line 826: email is guarded above?

nwingt (approved):

PR #1069, opened 2025-09-18
[TAS-4568] 👔 Update book sales email

williamchong:
Base on https://github.com/likecoin/likecoin-api-public/pull/1065

### DIFF
--- a/src/util/api/likernft/book/cart.ts
+++ b/src/util/api/likernft/book/cart.ts
@@ -39,7 +39,6 @@ import {
   sendNFTBookCartGiftPendingClaimEmail,
   sendNFTBookCartPendingClaimEmail,
   sendNFTBookOutOfStockEmail,
-  sendNFTBookSalesEmail,
 } from '../../../ses';
 import logPixelEvents from '../../../fbq';
 import { getBookUserInfoFromWallet } from './user';
@@ -50,7 +49,6 @@ import {
 import {
   CartItem, CartItemWithInfo, ItemPriceInfo, TransactionFeeInfo,
 } from './type';
-import { isEVMClassId } from '../../../evm/nft';
 import { isLikeNFTClassId } from '../../../cosmos/nft';
 
 export async function createNewNFTBookCartPayment(cartId: string, paymentId: string, {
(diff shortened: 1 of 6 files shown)


src/util/api/likernft/book/purchase.ts
```diff
@@ -156,7 +159,9 @@ export async function handleStripeConnectedAccount({
             currency,
             timestamp: FieldValue.serverTimestamp(),
           });
-          const shouldSendNotificationEmail = email && isEmailVerified;
+          const shouldSendNotificationEmail = !isOwner
+            && email
+            && isEmailVerified;
```
nwingt:
```suggestion
          const shouldSendNotificationEmailToChannel = !isOwner
            && email
            && isEmailVerified;
```

src/util/api/likernft/book/purchase.ts
```diff
@@ -257,7 +256,8 @@ export async function handleStripeConnectedAccount({
               email,
               isEmailVerified,
             } = likerUserInfo || {};
-            const shouldSendNotificationEmail = email && isEmailVerified;
+            const isOwner = wallet === ownerWallet;
+            const shouldSendNotificationEmail = !isOwner && email && isEmailVerified;
```
nwingt:
```suggestion
            const shouldSendNotificationEmailToConnectedAccount = !isOwner && email && isEmailVerified;
```

src/util/api/likernft/book/cart.ts
```diff
@@ -1158,6 +1159,16 @@ export async function formatCartItemInfosFromSession(session) {
       customPriceDiffInDecimal: acc.customPriceDiffInDecimal
         + item.customPriceDiffInDecimal * item.quantity,
       stripeFeeAmount: acc.stripeFeeAmount,
+      royaltyToSplit:
+        acc.royaltyToSplit
+        + (
+          item.priceInDecimal
+          - item.likerLandFeeAmount
```
nwingt, line 1167: is it intended not including `stripeFeeAmount`? Should we add `Math.max(0, ...)` like above?

williamchong: we don't know `stripeFeeAmount` here yet, so you can see it is always `acc.stripeFeeAmount` in the line above

nwingt: you mean we dont know `stripeFeeAmount` per item, but we know `stripeFeeAmount` for the whole cart?

williamchong: acc.stripeFeeAmount is always set to 0 here, and is expected to be replaced later when we know the actual fee value

nwingt (approved):

PR #1079, opened 2025-09-29
[TAS-4620] ✨ Support locale for Stripe checkout

nwingt:


src/util/stripe.ts
```diff
@@ -52,4 +52,21 @@ export async function getStripeFeeFromCheckoutSession(session: Stripe.Checkout.S
   return stripeFee?.amount || 0;
 }
 
+export function normalizeLanguageForStripeLocale(
+  language?: string,
```
williamchong:
```suggestion
  language: string | undefined,
```
Seems `language` could be `undefined` but is not really optional param

williamchong (changes requested):

PR #1086, opened 2025-10-02
📈 Add intercom subscription update events

williamchong:


src/util/api/plus/index.ts
```diff
@@ -353,6 +367,84 @@ export async function createNewPlusCheckoutSession(
...
+    });
+  }
+
+  if (subscription.status === 'canceled' || subscription.status === 'unpaid') {
+    // Check if it was a trial cancellation using previousAttributes or trial_end
+    const wasTrial = previousAttributes?.status === 'trialing';
+    await userCollection.doc(likerId).update({
+      likerPlus: {
+        ...user.likerPlus,
+        currentPeriodEnd: Date.now(),
```
nwingt: should be the end of current period instead of now?

williamchong: Actually only need to update this if `customer.subscription.deleted`, maybe I explicitly add a check

williamchong: Removed `customer.subscription.deleted` case, and just rely on `customer.subscription.updated` with check instead

nwingt (approved):

PR #1090, opened 2025-10-06
[TAS-4369] ✨ Add book approve slack cmd

williamchong:


src/routes/slack/book.ts
```diff
@@ -0,0 +1,148 @@
...
+        case 'help': {
+          res.json({
+            response_type: 'ephemeral',
+            text: `\`/book approve <classId> <approve_with_ads|approve_no_ads|reject>\` Approve or reject a book listing
+
+Examples:
+  \`/book approve likenft1abc123\` - Approve for listing & ads (default)
+  \`/book approve likenft1abc123 approve_with_ads\` - Approve for listing & ads
+  \`/book approve likenft1abc123 approve_no_ads\` - Approve for listing (no ads)
+  \`/book approve likenft1abc123 reject\` - Reject/hide listing`,
```
nwingt: should use `0x1234...5678` instead of `likenft1abc`?

src/routes/slack/book.ts
```diff
@@ -0,0 +1,148 @@
...
+      };
+      break;
+    default:
+      throw new Error(`Invalid approval action: ${action}`);
+  }
+
+  await likeNFTBookCollection.doc(classId).update(approvalUpdate);
+  await updateAirtablePublicationRecord({
+    id: classId,
+    ...approvalUpdate,
```
nwingt: need to create new fields?

williamchong: I think we just need to update `isHidden`, can change to explicit field

nwingt (approved):

PR #1093, opened 2025-10-08
🥅 Catch getEVMNftClassDataById err and support redirectClassId in cart

williamchong:


src/util/api/likernft/book/cart.ts
```diff
@@ -974,9 +974,9 @@ export async function formatCartItemsWithInfo(items: CartItem[]) {
       ]);
       if (!bookInfo) throw new ValidationError('NFT_NOT_FOUND');
       if (!metadata) throw new ValidationError('NFT_NOT_FOUND');
-      const { evmClassId } = bookInfo;
-      if (evmClassId && isLikeNFTClassId(classId)) {
-        classId = evmClassId as string;
+      const { evmClassId, redirectClassId } = bookInfo;
+      if (redirectClassId || (evmClassId && isLikeNFTClassId(classId))) {
```
nwingt, line 978: ???
```suggestion
      if (redirectClassId || (evmClassId && isLikeNFTClassId(evmClassId))) {
```

williamchong: i think the logic is to redirect to evmClassId only if classId is likenft1xxxx
For redirectClassId we dont need to check that

nwingt (approved):

PR #1114, opened 2025-11-20
🔥 Remove arweave v1 endpoint

williamchong:


nwingt (approved):
I pray to God that this isnt going to explode, Amen

PR #1134, opened 2025-12-12
[TAS-5055] 👔 Don't set plus coupon if checkout cart has external from channel

williamchong:


src/util/api/likernft/book/cart.ts
```diff
@@ -1307,6 +1308,11 @@ export async function handleNewCartStripeCheckout(inputItems: CartItem[], {
   cancelUrl?: string,
   language?: string,
 } = {}) {
+  let from: string = inputFrom as string || '';
+  if (!from || from === NFT_BOOK_DEFAULT_FROM_CHANNEL) {
+    from = NFT_BOOK_DEFAULT_FROM_CHANNEL;
+  }
```
nwingt: is the `from === NFT_BOOK_DEFAULT_FROM_CHANNEL` check redundant?

src/util/api/likernft/book/cart.ts
```diff
@@ -1341,7 +1347,8 @@ export async function handleNewCartStripeCheckout(inputItems: CartItem[], {
     const { email: userEmail, isEmailVerified, isLikerPlus } = likerUserInfo || {};
     customerId = bookUserInfo?.stripeCustomerId;
     customerEmail = isEmailVerified ? userEmail : email;
-    if (isLikerPlus) {
+
+    if (isLikerPlus && (!from || checkIsFromLikerLand(from)) && !coupon) {
```
nwingt: since `from` is always not empty, `!from` is redundant?

nwingt (approved):

PR #1137, opened 2025-12-16
⚡️ Cache NFT metadata for longer if it is migrated

williamchong:


src/routes/likerland/nft/metadata.ts
```diff
@@ -234,7 +235,9 @@ router.get('/nft/metadata', async (req, res, next) => {
       result.bookstoreInfo = bookstoreInfo;
     }
 
-    if (!hasAnyError) {
+    if (bookstoreInfo?.evmClassId || bookstoreInfo?.redirectClassId) {
+      res.set('Cache-Control', `public, max-age=${ONE_DAY_IN_S}, stale-while-revalidate=${ONE_DAY_IN_S}`);
```
nwingt, line 239: is this affecting approval state?

williamchong: No, since having `evmClassId` & `redirectClassId` mean this is a old class ID and should not be used

PR #1138, opened 2025-12-16
👔 Add approve_hidden status for approve slack cmd

williamchong:


src/routes/slack/book.ts
```diff
@@ -47,6 +47,15 @@ async function approveBook(classId: string, action: string, slackUserId: string)
         approvalStatus: 'approved_no_ads',
       };
       break;
+    case 'approve_hidden':
```
nwingt, line 50: why not just `hidden`? or `unlisted`

williamchong: Consistent with `approve_xx` like `approve_no_ads` and `approve_with_ads`, where `approve` is for sale and suffix is modifier (9up)

nwingt (approved):

PR #1139, opened 2025-12-17
⬆️ Upgrade Stripe SDK to clover

williamchong:


src/util/api/plus/index.ts
```diff
@@ -29,10 +29,10 @@ export async function processStripeSubscriptionInvoice(
 ) {
   const {
     billing_reason: billingReason,
-    discount,
-    subscription_details: subscriptionDetails,
+    parent,
   } = invoice;
-  const subscriptionId = invoice.subscription as string;
+  const subscriptionDetails = parent?.type === 'subscription_details' ? parent.subscription_details : null;
+  const subscriptionId = subscriptionDetails?.subscription as string;
```
nwingt, line 35: look likes `subscriptionId` is possibly `null`, not sure it is safe to cast it to string

williamchong: it is already checked before calling this function, but lets add a sanity check

nwingt (approved):

PR #1163, opened 2026-01-08
🧑‍💻 Use tsx to replace nodemon

williamchong:


src/util/airtable.ts
```diff
@@ -274,7 +276,7 @@ export async function updateAirtablePublicationRecord({
   isDRMFree,
   isHidden,
 }: UpdateAirtablePublicationRecordParams): Promise<void> {
-  if (!base) return;
+  if (!base) throw new Error('Airtable base is not initialized');
```
nwingt: Maybe refactor this?

nwingt (approved):

src/util/airtable.ts
williamchong: Maybe some function doesn't need to be exported, then we just need to guard those exported ones. Lets handle later

PR #1200, opened 2026-03-09
[TAS-4993] ✨ Add GA4 and posthog server events

williamchong:


src/util/logServerEvents.ts
nwingt, line 1: Please use kebab case filename

src/util/analyticsEvents.ts
nwingt, line 1: Please use kebab case filename

nwingt (approved):

PR #1206, opened 2026-03-11
[TAS-5427] ✨ Support multi-currency (USD/HKD/TWD) for Plus checkout and gift

williamchong:


src/routes/plus/index.ts
```diff
@@ -176,6 +183,7 @@ router.post('/gift/new', jwtAuth('write:plus'), async (req, res, next) => {
         period: period as 'monthly' | 'yearly',
         giftInfo,
         coupon,
+        currency: currency as 'usd' | 'hkd' | 'twd' | undefined,
```
nwingt: can refactor `['usd', 'hkd', 'twd']` and `'usd' | 'hkd' | 'twd'` type?

PR #1253, opened 2026-05-12
🐛 Emit deterministic uuid on transactional events for cross-source dedup

williamchong:
Pair with liker-land-v3's matching capture: the browser fires the same purchase/start_trial/subscribe/begin_checkout events from success pages after a user action. Both sides now build the same URL (`https://3ook.com/posthog-dedup/${posthogEvent}/${paymentId}`) and hash it under RFC 4122 NAMESPACE_URL to get a deterministic uuidv5, which is passed as the event uuid. PostHog's ClickHouse dedup tuple (timestamp, distinct_id, event, uuid) then collapses the pair into a single row. `$insert_id` is kept alongside as the ingest-layer belt to the new uuid's ClickHouse-layer suspenders.

The browser keys on the post-mapping event name, so the URL here must also use `posthogEvent` (after SERVER_EVENT_MAP), not the internal ServerEventName. Renaming a SERVER_EVENT_MAP value without updating liker-land-v3 will silently break dedup.

### DIFF
--- a/src/util/posthog.ts
+++ b/src/util/posthog.ts
@@ -1,4 +1,5 @@
 import { PostHog } from 'posthog-node';
+import uuidv5 from 'uuid/v5';
 
 import {
   POSTHOG_API_KEY,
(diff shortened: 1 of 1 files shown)


src/util/posthog.ts
```diff
@@ -7,6 +8,25 @@ import {
...
+//
+// The uuid is derived from the POST-MAPPING event name (`posthogEvent`, see capture call
+// below), NOT the internal `ServerEventName`. The browser passes the raw PostHog event
+// name (e.g. 'subscribe'), so we must match that side. Renaming a SERVER_EVENT_MAP value
+// without updating the browser would silently break dedup.
+const NAMESPACE_URL = '6ba7b811-9dad-11d1-80b4-00c04fd430c8'; // RFC 4122 appendix C
+
+// The host must be a fixed literal: substituting BOOK3_HOSTNAME or any other env-aware
+// hostname constant would flip between testnet/mainnet and silently break dedup.
+function derivePostHogEventUuid(eventName: string, transactionId: string): string {
```
nwingt:
```suggestion
function derivePostHogEventUUID(eventName: string, transactionId: string): string {
```

PR #1254, opened 2026-05-12
✨ Allow affiliate opt-in to share Plus 20% discount on @likerId links

williamchong:
Plus members arriving via `?from=@likerId` previously lost their 20% discount so the affiliate could earn the 30% channel commission. With the new `isPlusDiscountAllowed` flag (stored at the `NFTBookUserData` root because channel commission flows from `from` regardless of `affiliateConfig.active`), opted-in affiliates now let the discount apply while keeping `from` set — `calculateItemPrices` already deducts the discount from the channel commission, so the affiliate's 30% share absorbs the cost automatically.

Default is false, so existing affiliate links keep current behavior. The flag is admin-set in Firestore for now (no self-serve endpoint yet).

### DIFF
--- a/src/util/api/likernft/book/cart.ts
+++ b/src/util/api/likernft/book/cart.ts
@@ -52,7 +52,8 @@ import {
   sendPlusBookPromoCodeEmail,
 } from '../../../ses';
 import logServerEvents from '../../../logServerEvents';
-import { getBookUserInfoFromWallet } from './user';
+import { getBookUserInfoFromWallet, getBookUserInfoFromLikerId } from './user';
+import { normalizeLikerId } from '../../../ValidationHelper';
 import {
   SLACK_OUT_OF_STOCK_NOTIFICATION_THRESHOLD,
   LIKER_PLUS_20_COUPON_ID,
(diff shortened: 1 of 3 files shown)


AuroraHuang22 (approved):

src/util/api/likernft/book/cart.ts
```diff
@@ -1535,16 +1536,43 @@ export async function handleNewCartStripeCheckout(inputItems: CartItem[], {
...
-      itemInfos = itemInfos.map((item) => ({
-        ...item,
-        from: undefined,
-      }));
+    if (isLikerPlus && !coupon) {
+      let shouldApplyPlusDiscount = false;
+      // Default Liker Land channel: no real affiliate to pay — apply the
+      // discount and clear `from` so commission stays at zero.
+      let shouldClearFrom = false;
+      if (checkIsFromLikerLand(from)) {
```
nwingt:
```suggestion
      let shouldClearFrom = false;
      if (checkIsFromLikerLand(from)) {
        // Default Liker Land channel: no real affiliate to pay — apply the
        // discount and clear `from` so commission stays at zero.
```

nwingt (approved):

PR #1266, opened 2026-05-19
✨ Pre-warm shared ebook cache on book listing create and metadata sync

williamchong:
When a book listing is created or its on-chain metadata/file path is refreshed, stream the referenced files into the GCS bucket that the ebook-cors service reads from, so first reads hit a warm cache.

The cache key resolution mirrors ebook-cors byte-for-byte (arweave links resolved locally via Firestore) so entries are actually hit. Runs fire-and-forget so it never blocks the response or the ISCN sync.

### DIFF
--- a/src/util/api/likernft/book/cache.ts
+++ b/src/util/api/likernft/book/cache.ts
@@ -1,0 +1,194 @@
+import axios from 'axios';
+import { pipeline } from 'stream/promises';
+import type { File } from '@google-cloud/storage';
+
+import { ARWEAVE_GATEWAY, API_HOSTNAME } from '../../../../constant';
+import { bookCacheBucket } from '../../../gcloudStorage';
+import { getArweaveTxInfo } from '../../arweave/tx';
+import type { NFTClassData } from './index';
+
+// Mirrors the gateway lists in likecoin-cloud-functions/ebook-cors/nft/http.js.
+// The resolved URLs we cache always live on these gateways, so their cache time
+// is the permanent one (matching ebook-cors getUrlCacheTime()).
+const PERMANENT_CACHE_TIME_IN_S = 31536000;
+
+const LIKECOIN_API_LINK_PREFIX = `https://${API_HOSTNAME}/arweave/v2/link/`;
+
+// Arweave / IPFS gateway fallbacks, same ordering as ebook-cors.
+const ARWEAVE_GATEWAYS = ['https://arweave.net/', 'https://gateway.irys.xyz/'];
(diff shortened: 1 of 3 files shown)


src/util/api/likernft/book/cache.ts
```diff
@@ -0,0 +1,179 @@
...
+// Arweave / IPFS gateway fallbacks, same ordering as ebook-cors.
+const ARWEAVE_GATEWAYS = ['https://arweave.net/', 'https://gateway.irys.xyz/'];
+const IPFS_GATEWAYS = ['https://ipfs.io/ipfs/', 'https://w3s.link/ipfs/', 'https://gateway.irys.xyz/ipfs/'];
+
+function getCacheFilePath(classId: string, url: string) {
+  // Must match ebook-cors nft/cache.js getCacheFilePath() exactly.
+  return `${classId}/${encodeURIComponent(url)}`;
+}
+
+function getFallbackUrls(url: string): string[] {
```
nwingt: below is resolveBookFileCache`URL`

```suggestion
function getFallbackURLs(url: string): string[] {
```

src/util/api/likernft/book/cache.ts
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * The api/arweave/v2/link/<txHash> case is resolved locally via Firestore
+ * (getArweaveTxInfo) instead of an HTTP round-trip — it yields the same
+ * `${ARWEAVE_GATEWAY}/${arweaveId}` (+ ?key=) link the endpoint returns.
+ */
+async function resolveBookFileCacheURL(
+  targetURI: string,
+): Promise<string | undefined> {
+  if (!targetURI) return undefined;
+  let parsedUrl: URL;
```
nwingt: why targetURI but parsedURL 
```suggestion
  let parsedURL: URL;
```

src/util/api/likernft/book/cache.ts
```diff
@@ -0,0 +1,179 @@
...
+    const [metadata] = await fileRef.getMetadata();
+    // ebook-cors only treats an entry as a hit if contentType is set.
+    return !!metadata.contentType;
+  } catch (err) {
+    if ((err as { code?: number }).code === 404) return false;
+    throw err;
+  }
+}
+
+async function fetchStreamWithFallback(url: string, fallbackUrls: string[]) {
```
nwingt:
```suggestion
async function fetchStreamWithFallback(url: string, fallbackURLs: string[]) {
```

nwingt (approved):

PR #1273, opened 2026-05-22
[TAS-5899] 🐛 Reflect HKD/TWD price overrides in Airtable publication record

williamchong:
Min/Max Price in Airtable was computed purely from USD priceInDecimal, so a custom TWD/HKD override on a book tier never showed up. Emit per-currency "Min Price <CODE>" / "Max Price <CODE>" fields using the same override-aware resolver that drives Stripe currency_options.

The original USD columns stay untouched. Adding a new override currency to BOOK_PRICE_OVERRIDE_CURRENCIES now produces the matching Airtable fields automatically.

### DIFF
--- a/src/util/pricing.ts
+++ b/src/util/pricing.ts
@@ -61,6 +61,30 @@ export function getStripeCurrencyOptionsFromNFTBookPrice(
   );
 }
 
+type OverrideCurrency = typeof BOOK_PRICE_OVERRIDE_CURRENCIES[number];
+
+type BookPriceLike = {
+  priceInDecimal: number;
+  priceInDecimalByCurrency?: BookPriceInDecimalByCurrency;
+};
+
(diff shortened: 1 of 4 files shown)


src/util/pricing.ts
```diff
@@ -61,6 +61,30 @@ export function getStripeCurrencyOptionsFromNFTBookPrice(
   );
 }
 
+type OverrideCurrency = typeof BOOK_PRICE_OVERRIDE_CURRENCIES[number];
+
+type BookPriceLike = {
```
nwingt, line 66: What is Like in BookPriceLike mean?

nwingt (approved):

src/util/pricing.ts
williamchong: `BookPrice-Like` , sth shaped like a `BookPrice`

PR #1275, opened 2026-05-22
[TAS-5896] ✨ Bridge RevenueCat mobile IAP into Liker Plus subscriptions

williamchong:
Add an additive RevenueCat layer alongside the existing Stripe path so App/Play Store purchases grant Liker Plus, keyed by our internal user id as the RevenueCat app_user_id.

- POST /plus/revenuecat/webhook: auth via shared secret, maps RC events (purchase/renewal/expiration/billing issue/transfer) onto likerPlus. Ignores store=STRIPE events so the Stripe webhook stays the web owner.
- GET /plus/revenuecat/config: exposes the canonical app_user_id for the mobile app to call Purchases.logIn().
- Tag likerPlus with provider (stripe|revenuecat); shared record is latest-write-wins, terminal RC events won't revoke a Stripe-owned one.
- Best-effort POST /v1/receipts on Stripe subscription creation to unify web and mobile entitlements under one app_user_id.
- RevenueCat config keys are env-driven; tests set them via process.env since the config vi.mock does not reach source imports.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -51,9 +51,9 @@ config.LIKER_PLUS_GIFT_YEARLY_PRICE_ID = '';
 config.LIKER_PLUS_BOOK_PROMO_COUPON_CODE = '';
 
 // RevenueCat (mobile IAP) — bridges App/Play Store Plus subscriptions.
-// Product-id lists stay raw comma-separated strings here; the consumer parses them.
 config.REVENUECAT_WEBHOOK_AUTHORIZATION = process.env.REVENUECAT_WEBHOOK_AUTHORIZATION || ''; // shared secret set in RC dashboard's webhook Authorization header
 config.REVENUECAT_PLUS_ENTITLEMENT_ID = process.env.REVENUECAT_PLUS_ENTITLEMENT_ID || 'plus'; // RC entitlement identifier that grants Liker Plus
+// Product-id lists stay raw comma-separated strings here; the consumer parses them.
 config.REVENUECAT_PLUS_MONTHLY_PRODUCT_IDS = process.env.REVENUECAT_PLUS_MONTHLY_PRODUCT_IDS || ''; // comma-separated store product ids for the monthly period
 config.REVENUECAT_PLUS_YEARLY_PRODUCT_IDS = process.env.REVENUECAT_PLUS_YEARLY_PRODUCT_IDS || ''; // comma-separated store product ids for the yearly period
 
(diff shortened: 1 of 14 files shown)


config/config.js
```diff
@@ -50,6 +50,13 @@ config.LIKER_PLUS_GIFT_MONTHLY_PRICE_ID = '';
 config.LIKER_PLUS_GIFT_YEARLY_PRICE_ID = '';
 config.LIKER_PLUS_BOOK_PROMO_COUPON_CODE = '';
 
+// RevenueCat (mobile IAP) — bridges App/Play Store Plus subscriptions.
+// Product-id lists stay raw comma-separated strings here; the consumer parses them.
```
nwingt: should we put `// Product-id lists stay raw comma-separated strings here; the consumer parses them.` in line 57?

src/util/slack.ts
```diff
@@ -195,7 +196,11 @@ export async function sendPlusSubscriptionSlackNotification({
     const userLink = userId ? `<https://${LIKER_LAND_HOSTNAME}/${userId}|${userId}>` : 'N/A';
     const stripeEnvironment = IS_TESTNET ? 'test' : '';
     const customerLink = stripeCustomerId ? `<https://dashboard.stripe.com/${stripeEnvironment}/customers/${stripeCustomerId}|${stripeCustomerId}>` : 'N/A';
-    const subscriptionLink = `<https://dashboard.stripe.com/${stripeEnvironment}/subscriptions/${subscriptionId}|${subscriptionId}>`;
+    // Only Stripe ids resolve to a Stripe dashboard URL. Other providers (e.g.
+    // RevenueCat) pass their own transaction id, so render it as plain text.
+    const subscriptionLink = method === 'stripe'
+      ? `<https://dashboard.stripe.com/${stripeEnvironment}/subscriptions/${subscriptionId}|${subscriptionId}>`
```
nwingt, line 202: i wonder this can be rendered as link

src/util/misc.ts
```diff
@@ -20,3 +20,7 @@ export function maskString(
   if (!str || str.length <= start + end) return str;
   return `${str.slice(0, start)}*****${str.slice(-end)}`;
 }
+
+export function splitEnvList(value?: string): string[] {
```
nwingt: without context, does splitByComma make more sense?

nwingt (commented):
no data send to Airtable?

src/util/slack.ts
williamchong: probably not in this kind of formatting, but fix later

nwingt (approved):

PR #1277, opened 2026-05-26
✨ Validate path params and query strings with zod schemas

williamchong:
Extend zod validation from request bodies to path params and query strings across the arweave, likerland, likernft/book, plus, users and wallet surfaces. Adds validateParams to every route with a path param, and validateQuery where it tightens input the handler did not already guard.

Query schemas use .passthrough() and keep handler-guarded fields optional so existing error responses are preserved; priceIndex path segments are coerced to non-negative integers.


nwingt (approved):

nwingt:
@prayrequest

PR #1286, opened 2026-05-29
✨ Handle publisher object with description in book metadata

williamchong:
Publisher metadata can now be a { name, description } object like author. Add getNameFromMetadata helper and extract the publisher name for Stripe product metadata, Meta catalog brand, and Airtable.


nwingt (approved):

williamchong:
@prayrequest

PR #1297, opened 2026-06-04
[TAS-5912] ✨ Accrue Plus subscription daily value for reading revenue share

williamchong:
Compute a per-day value for each active non-trial Plus term — the funding basis for the upcoming reading-library revenue-share pool. dailyValue divides the actual net charge by the real term length (accurate to the day), so the early/beta/full price tiers and the monthly/yearly discount are captured without a price table.

Wired into both providers: Stripe invoices (in the invoice currency) and RevenueCat purchases/renewals (in USD, RevenueCat's normalized price). Trials contribute 0. Stored on the likerPlus record as dailyValue/dailyValueCurrency.

### DIFF
--- a/test/unit/plus-revenue-share.test.ts
+++ b/test/unit/plus-revenue-share.test.ts
@@ -1,0 +1,68 @@
+import { describe, it, expect } from 'vitest';
+import { ONE_DAY_IN_MS } from '../../src/constant';
+import { calculatePlusDailyValue } from '../../src/util/api/plus/revenueShare';
+
+const monthStart = Date.UTC(2026, 0, 1);
+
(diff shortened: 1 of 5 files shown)


test/unit/plus-revenue-share.test.ts
```diff
@@ -0,0 +1,65 @@
+import { describe, it, expect } from 'vitest';
+import { calculatePlusDailyValue } from '../../src/util/api/plus/revenueShare';
+
+const ONE_DAY = 86400000;
```
nwingt: reuse ONE_DAY_IN_MS?

nwingt (approved):

PR #1304, opened 2026-06-09
🔒️ Filter PII from unauthenticated EVM migrate lookup

williamchong:
The /wallet/evm/migrate/user/addr/:likeWallet endpoint returned the raw user doc (email, Stripe customerId/subscriptionId) without auth. Wrap likerIdInfo in filterUserDataMin so only public-safe fields leave, matching /users/addr/:addr/min.

Special thanks to @codevispera (Vincent C. from CodeVispera) for responsibly disclosing the high-severity unauthenticated PII issue addressed in this PR. Their professional report allowed us to identify and patch the vulnerability on the exact same day. Appreciate their efforts in helping us keep the platform secure.

### DIFF
--- a/src/routes/wallet/index.ts
+++ b/src/routes/wallet/index.ts
@@ -26,7 +26,9 @@ import {
 import { validateBody, validateParams } from '../../middleware/validate';
 import publisher from '../../util/gcloudPub';
 import { PUBSUB_TOPIC_MISC } from '../../constant';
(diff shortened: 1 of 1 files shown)


williamchong:
Just wanted to add a public note of thanks to @codevispera for the responsible disclosure that led to this fix, which allowed us to get this patched on the same day. 

Appreciate the professionalism and help in securing the ecosystem.

PR #1333, opened 2026-06-17
🐛 Tolerate legacy partial entries in affiliate config response

williamchong:
## Problem

`GET /plus/affiliate/:likerId` was 500ing with `RESPONSE_SCHEMA_MISMATCH` on real data:

```
RESPONSE_SCHEMA_MISMATCH: [{"code":"invalid_type","expected":"string","received":"undefined","path":["giftBooks",1,"classId"],"message":"Required"}]
```

Legacy/partial affiliate records in Firestore can hold gift books without a `classId`, or custom voices missing required string fields (`id`, `name`, `providerVoiceId`). The response schema required those fields, so a **single** malformed entry failed validation in `sendValidatedJSON` and took down the entire endpoint — not just the bad entry.

## Fix

Support the legacy data instead of dropping it: make those fields `.optional()` in `AffiliateGiftBookSchema` / `AffiliateCustomVoiceSchema` so partial entries pass response validation and are forwarded as-is.

- `priceIndex` stays required — the response builder always coerces it with `|| 0`, so the validated response always has a number.
- These schemas are `z.infer` sources for `AffiliateConfig` / `AffiliateGiftBook` / `AffiliateCustomVoice`, so the inferred types widen in lockstep. The only internal consumer that needed adjusting is the promo-voice mapping in `cart.ts`, which now filters out voices with no `name`.

## Consumer compatibility

The `3ook-com` BFF already tolerates partial upstream data — it types `customVoices` as `Partial<AffiliateCustomVoice>[]` and runtime-filters both arrays (`isValidCustomVoice`; `classId` presence for gift books). It keeps what it can render and ignores the rest, so forwarding legacy entries is safe end-to-end.

## Changes

- `src/util/api/plus/schemas.ts` — widen affiliate gift-book / custom-voice fields to optional
- `src/util/api/likernft/book/cart.ts` — guard promo-voice mapping against the now-optional `name`

### DIFF
--- a/src/util/api/likernft/book/cart.ts
+++ b/src/util/api/likernft/book/cart.ts
@@ -517,10 +517,12 @@ export async function processNFTBookCart(
           promoOwnerDisplayName = ownerLikerInfo.displayName;
           const ownerAffiliateConfig = ownerInfo?.bookUserInfo?.affiliateConfig;
           if (ownerAffiliateConfig?.active && ownerAffiliateConfig.customVoices?.length) {
-            promoOwnerVoices = ownerAffiliateConfig.customVoices.map((v) => ({
-              name: v.name,
-              language: v.language,
-            }));
+            promoOwnerVoices = ownerAffiliateConfig.customVoices
+              .filter((v): v is typeof v & { name: string } => !!v.name)
+              .map((v) => ({
+                name: v.name,
+                language: v.language,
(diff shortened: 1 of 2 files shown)


nwingt (approved):

williamchong:
@nwingt please review again

PR #1441, opened 2026-08-01
[TAS-6274] ✨ Support multi-month Plus gifts via monthly price quantity

nwingt:


src/util/api/plus/gift.ts
```diff
@@ -124,7 +127,7 @@ export async function createPlusGiftCheckoutSession(
     line_items: [
       {
         price: isYearly ? LIKER_PLUS_GIFT_YEARLY_PRICE_ID : LIKER_PLUS_GIFT_MONTHLY_PRICE_ID,
-        quantity: 1,
+        quantity: isYearly ? 1 : quantity,
```
williamchong: Actually can we have multiple year gift?

williamchong (approved):

PR #1449, opened 2026-08-06
[TAS-6292] ✨ Flag the claiming EVM wallet on book purchase responses

williamchong:


williamchong:
Wrong context, reverted

PR #1451, opened 2026-08-07
✨ Verify email by clicking the link, not via like.co

williamchong:
The verification mail linked to a like.co page that POSTed to the API, so the flow depended on a frontend we are shutting down. Add a GET route that verifies and renders a small self-contained result page instead, and build the link in code from API_EXTERNAL_HOSTNAME so testnet mail stops pointing at production.

Success no longer deletes verificationUUID: a mail scanner prefetching the link would otherwise burn the token before the reader clicks it. Because the UUID now outlives a verify, /users/update clears it when it resets isEmailVerified - previously a link mailed to an old address could verify the new one.

Also fixed along the way:
- /email/verify/user/:id sent 404 then fell through to 200
- a missing referral doc failed the whole verify; it is best-effort now
- drop the dead SES copy of sendVerificationEmail


src/util/api/email/verifyPage.ts
```diff
@@ -0,0 +1,81 @@
...
+<title>${title}</title>
+<style>
+:root { color-scheme: light dark; }
+body {
+  margin: 0;
+  min-height: 100vh;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background: #f7f7f5;
```
nwingt, line 37:
```suggestion
  background: #f9f9f9;
```

src/util/api/email/verifyPage.ts
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable no-underscore-dangle */
+import type { Response } from 'express';
+import { getBook3URL } from '../../liker-land';
+
+export type EmailVerifyPageStatus = 'success' | 'failed' | 'error';
+
+const BRAND_COLOR = '#28646E';
```
nwingt, line 7: use `#131313` instead of LikeCoin green?

src/util/api/email/verifyPage.ts
```diff
@@ -0,0 +1,81 @@
...
+  background: ${isSuccess ? BRAND_COLOR : '#9b9b9b'};
+}
+h1 { font-size: 1.375rem; margin: 0 0 0.5rem; }
+p { margin: 0 0 2rem; color: #4a4a4a; }
+a.cta {
+  display: inline-block;
+  padding: 0.75rem 1.75rem;
+  border-radius: 0.375rem;
+  background: ${BRAND_COLOR};
+  color: #fff;
```
nwingt, line 62:
```suggestion
  color: #f9f9f9;
```

src/util/api/email/verifyPage.ts
```diff
@@ -0,0 +1,81 @@
...
+  text-decoration: none;
+  font-weight: 600;
+}
+@media (prefers-color-scheme: dark) {
+  body { background: #16181a; color: #f2f2f2; }
+  p { color: #b8b8b8; }
```
nwingt, line 68:
```suggestion
  body { background: #131313; color: #f2f2f2; }
  p { color: #f9f9f9; }
```

src/util/api/email/verifyPage.ts
```diff
@@ -0,0 +1,81 @@
...
+<style>
+:root { color-scheme: light dark; }
+body {
+  margin: 0;
+  min-height: 100vh;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background: #f7f7f5;
+  color: #1a1a1a;
```
nwingt, line 38:
```suggestion
  color: #131313;
```

PR #1462, opened 2026-08-12
✨ Let users rename their Liker ID without moving the account key

williamchong:
The `user` field was doing six jobs at once: Firestore document id, JWT subject, Intercom external id, RevenueCat appUserID, OAuth subject, and public handle. Only the last one wants to be mutable.

Keep `user` meaning exactly what it meant — the immutable document id — and add `handle` as the mutable public name, resolved through a handle -> id index. Nothing downstream changes meaning, so the mobile app's `likerId == document id` invariant and 3ook's session cookie stay valid with no coordinated deploy.

Old handles become permanent aliases rather than being freed: they still resolve to their owner and can never be re-registered, so `from=@handle` links already in circulation keep crediting the right account.

Also stores Plus affiliate attribution as the internal id, fixing an existing bug where `/plus/affiliate` compared a stored handle against an internal id and could list one affiliate twice.

See docs/adr/0003-liker-id-handle-split.md.


src/util/api/users/handle.ts
```diff
@@ -0,0 +1,165 @@
...
+      t.get(newIndexRef),
+      t.get(newHandleUserRef),
+    ]);
+    const userData = userDoc.data();
+    if (!userDoc.exists || !userData || userData.isDeleted) {
+      throw new ValidationError('USER_NOT_FOUND', 404);
+    }
+    const currentHandle = getUserHandle(userId, userData);
+    if (userData.handle) throw new ValidationError('HANDLE_ALREADY_CHANGED', 409);
+    if (currentHandle === newHandle) throw new ValidationError('HANDLE_UNCHANGED', 400);
```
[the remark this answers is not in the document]
nwingt: seems valid

src/util/api/users/handle.ts
```diff
@@ -0,0 +1,171 @@
...
+  return userData?.handle || userId;
+}
+
+/**
+ * Whether `requestedId` — a handle, an alias, or an internal id from a route
+ * param — names the same account as the internal id in the caller's token. The
+ * token always carries the internal id, so the equality fast path covers every
+ * request that did not address the user by handle.
+ */
+export async function isSelfUserId(
```
nwingt: the function name feels off, how about `isSameUser`?

nwingt: the function name feels off, how about `isSameUser`?

src/util/api/users/handle.ts
```diff
@@ -0,0 +1,171 @@
...
+  claimedUserId: string,
```
nwingt:
```suggestion
  authenticatedUserId: string,
```

src/util/api/users/handle.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  // Document id first: until an account renames, its handle *is* its document id,
+  // so the overwhelmingly common case costs the single read it always did. Probed
+  // as written before lowercased, because document ids are case-sensitive and
+  // legacy ids predate the lowercase-only rule.
+  const casings = lowered === handle ? [handle] : [handle, lowered];
+  for (const candidate of casings) {
+    // eslint-disable-next-line no-await-in-loop
+    const selfDoc = await userCollection.doc(candidate).get();
+    if (selfDoc.exists) {
```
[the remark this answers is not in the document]
nwingt: no mixed-case id in the past

